PR #16, opened 2021-03-23
Add thermal model

teubert:
Add Thermal Model from Chetan Kulkarni


chetankul (approved):
Updates to the thermal model are integrated correctly into the battery model.

PR #30, opened 2021-05-10
Feature/derived params

teubert:
Added functionality for derived parameters that automatically update when the parameters they are depends on change. Each model can implement a function that defines the derived parameter callbacks for that model.


jason-watkins (approved):
Aside from the LGTM warning, looks good to me.

PR #32, opened 2021-05-11
Added Simple Degradation to Electrochem Battery model

teubert:
Added simple degradation in capacity to the electrochemical model. Now there are three models:

- `BatteryElectroChemEOD`: standard discharge model
- `BatteryElectroChemEOL`: capacity degradation model
- `BatteryElectroChemEODEOL`: Combined model


chetankul (approved):
All looks good. I very much like the way classes are defined individually as well as combined.

Issue #43, opened 2021-05-26
Improve stability of BatteryElectroChemEOD Model


teubert:
BatteryElectroChemEOD model can lead to unstable states. In these cases 

File "/Users/cteubert/dev/prog_models/src/prog_models/models/battery_electrochem.py", line 321, in output
```Python
    Vep = params['U0p'] + R*x['tb']/F*log((1-xpS)/xpS) + sum(VepParts)
```
ValueError: math domain error

In this case `log((1-xpS)/xpS)` reaches -infinity, where xpS = x['qpS']/params['qSMax']. This happens because `qpS` approaches qSMax. 

I wonder if we can update the model in such a way that will prevent this state. For example:
```Python
qpSdot = max(qpSdot, (params['qSMax']-x['qpS'])/dt)
```

The problem with this is that we would have to switch from using dx to next_state (since dx doesn't have an idea of what dt is. Another option could be to add a range-checking feature for the states. This would slow down execution though to do these checks every step of the way. 

Another consideration is that we don't want to arbitrarily cap these states. We would want to make sure that the resulting behavior accurately represents battery behavior in this case. I want to dig deeper into what this unstable state is. 

@chetankul - Will need your help with this

PR #45, opened 2021-06-03
Fix bug in SOC Calculation

teubert:
See #43 

Battery EOD model becomes unstable when using SOC to maintain battery change (when simulating to EOD). There are cases where SOC as calculated is >0 but EOD has been met, leading to the above approach simulating past EOD. 

We have never run into this in the past as we've never used SOC to regulate future loading


chetankul (approved):
This section of the updated code reads good. An as per the tests performed the error is not observed.

PR #51, opened 2021-06-06
Pump Improvements

### LINKED ISSUE #50: Split CentrifugalPump into parts
Split CentrifugalPump into CentrifugalPumpBase and CentrifugalPumpWithWear, with CentrifugalPump defaulting to CentrifugalPumpWithWear. CentrifugalPumpBase is same as current CentrifugalPump, but without the wear parameters

teubert:
The following improvements:
* Added example of using pump. 
* Fixed bug where pump event states were negative
* Split model into base and with wear
* Add .vscode to .gitignore

Closes #50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,5 @@ cython_debug/
 .pypirc
 
 ex.pickle
+
+.vscode/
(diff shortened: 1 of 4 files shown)


chetankul (approved):
As discussed the wear parameters are not part of state space. The code reads good to me.

PR #55, opened 2021-06-15
Valve example

wlamberson1:


examples/sim_valve.py
```diff
@@ -0,0 +1,60 @@
...
+    config = {'dt': 0.01, 'horizon': 800, 'save_freq': 60}
+    # set wear parameter for spring to 1
+    valv.parameters['x0']['wk'] = 1
+
+    (times, inputs, states, outputs, event_states) = valv.simulate_to_threshold(future_loading, valv.output(valv.initialize(future_loading(0))), **config)
+
+    # Print Results
+    for i in range(len(times)):
+        print("Time: {}\n\tInput: {}\n\tState: {}\n\tOutput: {}\n\tEvent State: {}\n".format(times[i], inputs[i], states[i], outputs[i], event_states[i]))
+
```
teubert, line 44: Could you add a comment here. Something like:
```Python
# Now lets repeat the same test with a different wear mode. 
```

examples/sim_valve.py
```diff
@@ -0,0 +1,60 @@
...
+    print('\n\n------------------------------------------------')
+    print('Simulating to threshold\n\n')
+    config = {'dt': 0.01, 'horizon': 800, 'save_freq': 60}
+    # reset wear parameter for spring to 0, set wear parameter for friction to 1
+    valv.parameters['x0']['wk'] = 0
+    valv.parameters['x0']['wr'] = 1
+
+    (times, inputs, states, outputs, event_states) = valv.simulate_to_threshold(future_loading, valv.output(valv.initialize(future_loading(0))), **config)
```
teubert, line 52: I know this is what we do in the test, but for readability in the example, let's pull the output(initialize(future_loading(0))) to a different line and set it to variable first_output

examples/sim_valve.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Print Results
+    for i in range(len(times)):
+        print("Time: {}\n\tInput: {}\n\tState: {}\n\tOutput: {}\n\tEvent State: {}\n".format(times[i], inputs[i], states[i], outputs[i], event_states[i]))
+
+# This allows the module to be executed directly
+if __name__ == '__main__':
+    run_example()
```
teubert, line 60: Could you add a newline to the bottom of the file. Some of the other files I worked on need that too, but we can fix that in another PR. 

It's not strictly needed, but it's standard practice

teubert (changes requested):
Make sure you also add the other changes from the linked issue to this PR (e.g., adding example to tests, documentation, etc.)

teubert (approved):

PR #70, opened 2021-06-29
Automatic Noise Generation

wlamberson1:
Noise is automatically generated in private functions __next_state and __output functions

### DIFF
--- a/examples/new_model.py
+++ b/examples/new_model.py
@@ -42,7 +42,8 @@ class ThrownObject(PrognosticsModel):
             }
     
     def dx(self, x, u):
-        return {'x': x['v'], 'v': self.parameters['g']} # g is the acceleration of gravity
+        return {'x': x['v'],
+                'v': self.parameters['g']} # Acceleration of gravity
 
     def output(self, x):
         return {'x': x['x']}
--- a/examples/new_model.py
+++ b/examples/new_model.py
@@ -42,16 +42,10 @@ class ThrownObject(PrognosticsModel):
             }
     
     def dx(self, x, u):
-        # apply_process_noise is used to add process noise to each step
-        return self.apply_process_noise({
-            'x': x['v'],
-            'v': self.parameters['g'] # Acceleration of gravity
-        })
+        return x
 
     def output(self, x):
-        return self.apply_measurement_noise({
-            'x': x['x']
-        })
+        return x
 
     # This is actually optional. Leaving thresholds_met empty will use the event state to define thresholds.
     #  Threshold = Event State == 0. However, this implementation is more efficient, so we included it
--- a/examples/new_model.py
+++ b/examples/new_model.py
@@ -42,10 +42,10 @@ class ThrownObject(PrognosticsModel):
             }
     
     def dx(self, x, u):
-        return x
+        return {'x': x['v'], 'v': self.parameters['g']} # g is the acceleration of gravity
 
     def output(self, x):
-        return x
+        return {'x': x['x']}
 
     # This is actually optional. Leaving thresholds_met empty will use the event state to define thresholds.
     #  Threshold = Event State == 0. However, this implementation is more efficient, so we included it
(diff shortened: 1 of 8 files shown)


examples/new_model.py
```diff
@@ -42,16 +42,10 @@ def initialize(self, u, z):
             }
     
     def dx(self, x, u):
-        # apply_process_noise is used to add process noise to each step
-        return self.apply_process_noise({
-            'x': x['v'],
-            'v': self.parameters['g'] # Acceleration of gravity
-        })
+        return x
```
teubert, line 45: Here you're returning the state (x, v) instead of the first derivative of state (v, a | a=g).

You should just be removing the call to apply_process_noise

examples/new_model.py
```diff
@@ -42,16 +42,10 @@ def initialize(self, u, z):
...
 
     def output(self, x):
-        return self.apply_measurement_noise({
-            'x': x['x']
-        })
+        return x
```
teubert, line 48: Same as above, here you're returning state (x) instead of output (z = {'x': x['x'})

src/prog_models/prognostics_model.py
```diff
@@ -405,6 +405,51 @@ def next_state(self, x, u, dt) -> dict:
...
+        --------
+        next_state
+
+        Note
+        ----
+        A model should not overwrite '__next_state'
+        """
+        
+        # Calculate next state
+        self.next_state(x, u, dt)
```
teubert, line 448: Here you never actually save the next state. 

It should be
```suggestion
        X = self.next_state(x, u, dt)
```

src/prog_models/prognostics_model.py
```diff
@@ -458,6 +503,37 @@ def output(self, x) -> dict:
...
+        | z = {'z1': 2.2}
+        | x = m.initialize(u, z) # Initialize first state
+        | z = m.__output(3.0, x) # Returns {'o1': 1.2} with noise added
+        """
+
+        # Calculate next state, forward one timestep
+        self.output(x)
+
+        # Add measurement noise
+        return self.apply_measurement_noise(self.output(x))
```
teubert, line 535: Rather than calling self.output(x) twice, try saving it as a variable (z) on the first call

src/prog_models/prognostics_model.py
```diff
@@ -405,6 +405,51 @@ def next_state(self, x, u, dt) -> dict:
...
+        -------
+        | m = PrognosticsModel() # Replace with specific model being simulated
+        | u = {'u1': 3.2}
+        | z = {'z1': 2.2}
+        | x = m.initialize(u, z) # Initialize first state
+        | x = m.__next_state(x, u, 0.1) # Returns state, with noise, at 3.1 seconds given input u
+        
+        See Also
+        --------
+        next_state
```
teubert, line 440: This is very good that you refer to the original next-state

src/prog_models/prognostics_model.py
```diff
@@ -405,6 +405,51 @@ def next_state(self, x, u, dt) -> dict:
         dx = self.dx(x, u)
         return {key: x[key] + dx[key]*dt for key in dx.keys()}
 
+    def __next_state(self, x, u, dt) -> dict:
+        """
+        State transition equation: Calculate next state and add noise
```
teubert, line 410: Could you add a comment here explaining the difference between __next_state and next_state and how next_state is meant to be overwritten. 

Same with output, below

teubert (changes requested):
Overall is what I'm looking for, a few mistakes in implementation that need to be fixed before deployment.

examples/new_model.py
wlamberson1: so I should just return self instead of returning self.apply_process_noise()?

jason-watkins (approved):

teubert (approved):

examples/new_model.py
```diff
@@ -42,16 +42,10 @@ def initialize(self, u, z):
             }
     
     def dx(self, x, u):
-        # apply_process_noise is used to add process noise to each step
-        return self.apply_process_noise({
-            'x': x['v'],
-            'v': self.parameters['g'] # Acceleration of gravity
-        })
+        return {'x': x['v'], 'v': self.parameters['g']} # g is the acceleration of gravity
```
teubert, line 45: Actually, could you change this to be multiple lines for readability?

PR #74, opened 2021-07-15
Pneumatic Valve Split

wlamberson1:
Split PneumaticValve model into PneumaticValveBase and PneumaticValveWithWear (like pump model).  Updated tests to reflect changes.


jason-watkins (approved):

teubert (approved):

teubert:
Looks good. Good work!

teubert (changes requested):
Actually just caught something. Make sure you remove the apply_noise calls from within the class (since they're now done in the private classes)

teubert (approved):
Approve- we're opening up a separate issue to remove the apply_noise functions from every model

PR #80, opened 2021-07-26
Update prognostics_model.py

lizjhale:
Rename "observables" to "observables_keys" in model initialization to avoid conflict with function "observables"


wlamberson1 (approved):
Looks good to me

teubert (approved):

PR #81, opened 2021-07-26
Automated Testing of Individual Test Execution

wlamberson1:


teubert:
@wlamberson1 i know we talked about this method but it occurred to me that we should probably replace the way we call each set of tests to do this rather than calling the tests twice. Is there a way to do that and still have the executable fail if any tests fail?

tests/__main__.py
```diff
@@ -1,12 +1,14 @@
 # Copyright © 2020 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
+from cffi.recompiler import GlobalExpr
```
teubert:
```suggestion
```

tests/__main__.py
```diff
@@ -1,12 +1,14 @@
...
+from setuptools.sandbox import _execfile
```
teubert:
```suggestion
```

teubert (approved):

PR #105, opened 2021-08-10
Remove exec()

wlamberson1:


teubert (approved):

tests/__main__.py
```diff
@@ -31,32 +31,27 @@ def _test_ex():
     # Run tests individually to test them and make sure they can be executed individually
     was_successful = True
     try:
-        exec(open("tests/test_base_models.py").read())
+        base_models_main()
     except Exception:
         was_successful = False
 
     try:
-        exec(open("tests/test_sim_result.py").read())
```
jason-watkins, line 39: It looks like there's an uncaught conflict here. We seem to be deleting the sim_result tests

jason-watkins (changes requested):

tests/__main__.py
wlamberson1: I dont know where "test_sim_result" came from, but it is not one of the original tests.

jason-watkins: Yes, I believe it was added in #100, merged an hour ago.

jason-watkins (approved):

PR #106, opened 2021-08-10
Resolve doc comments

teubert:


wlamberson1 (approved):
Looks good. Maybe add something that clarifies what OS you're using when you give commands to run.
Windows: py -m tests
Mac:         python -m tests

PR #111, opened 2021-08-10
Optional State Bounds

wlamberson1:


teubert (changes requested):

teubert:
I looked at the existing models and here are the obvious bounds I see:

# Battery Circuit
* tb is a temperature in C, so [-273.15, inf]

# BatteryElectroChemEOD
* tb is a temperature in C, so [-273.15, inf]

# BatteryElectroChemEODEOL
* same as BatteryElectroChemEOD

# CentrifugalPumpBase
* To, Tr, Tt are temperatures in °C so [-273.15, inf]
* A is Area so [0, inf]

# CentrifugalPumpWithWear
* Same as CentrifugalPumpBase 

# PneumaticValveBase
* Aeb, Art, Ai are all areas - so [0, inf]
* k, mBot, mTop, r are all [0, inf]

# PneumaticValve with wear
* Same as base

src/prog_models/models/battery_circuit.py
```diff
@@ -3,6 +3,7 @@
 
 from .. import prognostics_model
 
+import math
```
teubert, line 6: to avoid double importing parts of math, let's remove this and change line 7 to from math import exp, inf. and then change math.inf -> inf. 

Same thing in the other files where parts of math are imported

wlamberson1: oh okay. I tried this but i left it as math.inf and it wasn't working

src/prog_models/models/pneumatic_valve.py
```diff
@@ -167,6 +167,16 @@ class PneumaticValveBase(prognostics_model.PrognosticsModel):
         'wt': 0
     }
 
+    state_limits = {
+        'Aeb': (0, inf),
+        'Art': (0, inf),
```
teubert, line 172:
```suggestion
        'Aet': (0, inf),
```

teubert (approved):

PR #113, opened 2021-08-11
Release v1.1

teubert:
Updates in V1.1
---------------
* New Feature: Derived Parameters
    * Users can specify callbacks for parameters that are defined from others. These callbacks will be called when the dependency parameter is updated.
    * See `examples.derived_params` for more information.
* New Feature: Parameter Estimation
    * Users can use the estimate_parameters method to estimate all or select parameters. 
    * see `examples.param_est`
* New Feature: Automatic Noise Generation
    * Now noise is automatically generated when next_state/dx (process_noise) and output (measurement_noise). This removed the need to explicitly call apply_*_noise functions in these methods. 
    * See `examples.noise` for more details in setting noise
    * For any classes users created using V1.0.*, you should remove any call to apply_*_noise functions to prevent double noise application. 
* New Feature: Configurable State Bounds
    * Users can specify the range of valid values for each state (e.g., a temperature in celcius would have to be greater than -273.15 - absolute zero)
* New Feature: Simulation Result Class
    * Simulations now return a simulation result object for each value (e.g., output, input, state, etc) 
    * These simulation result objects can be used just like the previous lists. 
    * Output and Event State are now "Lazily Evaluated". This speeds up simulation when intermediate states are not printed and these properties are not used
    * A plot method has been added directly to the class (e.g., `event_states.plot()`)
* New Feature: Intermediate Result Printing
    * Use the print parameter to enable printing intermediate results during a simulation 
    * e.g., `model.simulate_to_threshold(..., print=True)`
    * Note: This slows down simulation performance
* Added support for python 3.9
* Various bug fixes

ElectroChemistry Model Updates
==============================
* New Feature: Thermal effects to model - how the temperature changes with use. Previous implementation only included effects of temperature on performance.
* New Feature: Added `degraded_capacity` (i.e., EOL) event to model. There are now three different models: BatteryElectroChemEOL (degraded_capacity only), BatteryElectroChemEOD (discharge only), and BatteryElectroChemEODEOL (combined). BatteryElectroChem is an alias for BatteryElectroChemEODEOL. 
* New Feature: Updated SOC (EOD Event State) calculation to include voltage when near V_EOD. This prevents a situation where the voltage is below lower bound but SOC > 0. 
* Added ranges and descriptions for model states
* Various bug fixes

CentrifugalPump Model Updates
=============================
* New Feature: Added CentrifugalPumpBase class where wear rates are parameters instead of part of the state vector. 
    * Some users may use this class for prognostics, then use the parameter estimation tool occasionally to update the wear rates, which change very slowly.
* Bugfix: Fixed bug where some event states were returned as negative
* Bugfix: Fixed bug where some states were saved as parameters instead of part of the state. 
* Added example on use of CentrifugalPump Model (see `examples.sim_pump`)
* Added ranges and descriptions for model states
* Performance improvements

PneumaticValve Model Updates
============================
* New Feature: Added PneumaticValveBase class where wear rates are parameters instead of part of the state vector. 
    * Some users may use this class for prognostics, then use the parameter estimation tool occasionally to update the wear rates, which change very slowly.
* Added ranges and descriptions for model states
* Added example on use of PneumaticValve Model (see `examples.sim_valve`)

BatteryCircuit Model Updates
============================
* Added ranges and descriptions for model states


teubert:
Recorded the following comments in the release review:
1. Doc Navigation is broken
2. Make sure that qMobile callback works correctly in EOLEOD
3. EOLEOD - states just show eol state
5. ProgModel Template Example callback 
6. Example in estimate_params docs
7. May be missing test for param est. 
8. Test Param est. for (battery) - complex params  
9. Add process noise comment - line 456
10. Progress_bar - future feature
11. Newline at end of file - sim
12. Example of bounds
13. Chetan - limits of existing models
14. CachedSimResult -> LazySimResult
15. is_cached missing doc comments 
16. Added thermal effects to model - clear up
17 BatteryElectroChemEODEOL two lines before
18. Release notes- last line - engineers capitalized 
19. Release notes - Bug fixes

README.md
```diff
@@ -1,48 +1,52 @@
...
 `docs/` - Project documentation<br />
 `sphinx_config/` - Configuration for automatic documentation generation<br />
-`examples/` - Example python scripts using prog_models<br />
+`examples/` - Example Python scripts using prog_models<br />
 `tests/` - Tests for prog_models<br />
 `README.md` - The readme (this file)<br />
-`requirements.txt` - python library dependiencies required to be met to use this package. Install using `pip install -r requirements.txt`<br />
-`prog_model_template.py` - Template for Prognsotics Model<br />
+`requirements.txt` - Python library dependiencies required to be met to use this package<br />
+`prog_model_template.py` - Template for Prognsostics Model<br />
```
casonwight, line 30: "Prognostics"

src/prog_models/models/centrifugal_pump.py
```diff
@@ -121,79 +122,92 @@ class CentrifugalPump(prognostics_model.PrognosticsModel):
...
+        'Tt': (-273.15, inf),
+        'A': (0, inf)
+    }
+
     def initialize(self, u, z = None):
         x0 = self.parameters['x0']
-        x0['QLeak'] = math.copysign(self.parameters['cLeak']*self.parameters['ALeak']*math.sqrt(abs(u['psuc']-u['pdisch'])), u['psuc']-u['pdisch'])
+        x0['QLeak'] = math.copysign(\
+            self.parameters['cLeak']*self.parameters['ALeak']*\
+                math.sqrt(abs(u['psuc']-u['pdisch'])), u['psuc']-u['pdisch'])
```
casonwight, line 154: From the following paper, it seems like the sign should be reversed (i.e. u['pdisch']-u['psuc'] instead of u['psuc']-u['pdisch']):
Daigle, M. J., & Goebel, K. (2013). Model-based prognostics with concurrent damage progression processes. IEEE Transactions on Systems, Man, and Cybernetics: Systems, 43(3), 535–546. https://doi.org/10.1109/tsmca.2012.2207109

src/prog_models/models/centrifugal_pump.py
```diff
@@ -121,79 +122,92 @@ class CentrifugalPump(prognostics_model.PrognosticsModel):
...
-            'To': x['To'] + Todot*dt,
-            'Tr': x['Tr'] + Trdot*dt,
-            'Tt': x['Tt'] + Ttdot*dt,
-            'rRadial': x['rRadial'] + rRadialdot*dt,
-            'rThrust': x['rThrust'] + rThrustdot*dt,
-            'w': x['w']+wdot*dt,
-            'wA': x['wA'],
-            'wRadial': x['wRadial'],
-            'wThrust': x['wThrust'],
+        Te = 3*params['R2']/slip/(u['wsync']+0.00001)*u['V']**2 \
```
casonwight, line 175: Should the number of phases (n) and the number of pole pairs (p) be parameters of the class? On this line, 3 is hard coded, limiting model flexibility. See paper:
Daigle, M. J., & Goebel, K. (2013). Model-based prognostics with concurrent damage progression processes. IEEE Transactions on Systems, Man, and Cybernetics: Systems, 43(3), 535–546. https://doi.org/10.1109/tsmca.2012.2207109

src/prog_models/models/centrifugal_pump.py
```diff
@@ -121,79 +122,92 @@ class CentrifugalPump(prognostics_model.PrognosticsModel):
...
+            /((params['R1']+params['R2']/slip)**2+(u['wsync']*params['L1'])**2)
+        backTorque = -params['a2']*Qout**2 + params['a1']*x['w']*Qout + params['a0']*x['w']**2
+        Qo = math.copysign(params['c']*math.sqrt(abs(deltaP)), deltaP)
+        wdot = (Te-friction-backTorque)/params['I']
+        Qdot = 1/params['FluidI']*(Qo-x['Q'])
```
casonwight, line 180: From the proceeding paper, it looks like the sign may be flipped here (i.e. x['Q']-Qo instead of Qo-x['Q']):

Daigle, M. J., & Goebel, K. (2013). Model-based prognostics with concurrent damage progression processes. IEEE Transactions on Systems, Man, and Cybernetics: Systems, 43(3), 535–546. https://doi.org/10.1109/tsmca.2012.2207109

src/prog_models/prognostics_model.py
```diff
@@ -1,23 +1,43 @@
...
+    Args:
+        model: PrognosticsModel for which the params correspond
+        dict_in: Initial parameters
+        callbacks: Any callbacks for derived parameters f(parameters) : updates (dict)
     """
-    def __init__(self, model, dict_in = {}):
+    def __init__(self, model, dict_in = {}, callbacks = {}):
         super().__init__()
         self.__m = model
+        self.callbacks = {}
```
casonwight, line 28: The need for this line is not clear because line 35 resets the classes' callbacks (defaulting to {}). We did a small test and it doesn't seem to work without it though. This is more of a curiosity for us than a bug/fix.

teubert: This is needed to prevent the initial setting on 31 from calling the callbacks. This could call a callback which uses a parameter that hasn't yet been set, depending on parameter order as it iterates through dict_in on line 29.

teubert: I added a comment there to explain for other users. Excellent question!

src/prog_models/prognostics_model.py
```diff
@@ -342,7 +402,59 @@ def next_state(self, x, u, dt) -> dict:
...
+        ----
+        A model should not overwrite '__next_state'
+        A model should overwrite either `next_state` or `dx`. Override `dx` for continuous models, and `next_state` for discrete, where the behavior cannot be described by the first derivative.
+        """
+        
+        # Calculate next state and add process noise
+        next_state = self.apply_process_noise(self.next_state(x, u, dt))
+
+        # Check if state is within bounds
+        for (key, limit) in self.state_limits.items():
```
casonwight, line 452: This may not be feasible for this release, but it may make more sense to sample directly from a truncated distribution (that fits within the limits) instead of setting the state to the exact value of the limit. That way the limit itself is never reached by the state. This could be done with a modification on apply_process_noise.

src/prog_models/models/centrifugal_pump.py
teubert: Currently investigating this one. The implemented solution is consistent with the matlab implementation as well. Currently I suspect this might be a mistake in the paper, but we're working to confirm.

src/prog_models/models/centrifugal_pump.py
teubert: Good point- I'll open a PR shortly to implement this.

src/prog_models/models/centrifugal_pump.py
teubert: Currently investigating this one. The implemented solution is consistent with the matlab implementation as well. Currently I suspect this might be a mistake in the paper, but we're working to confirm.

src/prog_models/models/centrifugal_pump.py
teubert: Opened PR #120 Could you review that for me?

src/prog_models/models/battery_electrochem.py
```diff
@@ -164,16 +191,32 @@ class BatteryElectroChem(prognostics_model.PrognosticsModel):
...
-        'VEOD': 3.0
-    })
-
-    def initialize(self, u, z):
+        'VEOD': 3.0, 
+        'VDropoff': 0.1 # Voltage above EOD after which voltage will be considered in SOC calculation
+    }
+
+    state_limits = {
+        'tb': (-273.15, inf)
```
casonwight, line 205: The state is in Kelvin, while the output is in Celsius. The state_limits only affect the states, so it seems like the minimum temperature should also be in Kelvin (e.g. 0 instead of -273.15).

src/prog_models/models/centrifugal_pump.py
```diff
@@ -121,79 +122,92 @@ class CentrifugalPump(prognostics_model.PrognosticsModel):
...
-            'wA': 0.0,
-            'wThrust': 0,
-            'wRadial': 0,
         }
     }
 
+    state_limits = {
+        'To': (-273.15, inf),
+        'Tr': (-273.15, inf),
+        'Tt': (-273.15, inf),
```
casonwight, line 146: The state limits here seem like they should be in Kelvin to agree with the rest of the model (e.g. 0 instead of -273.15). If the intention is to have everything in the model be in Celsius, then x0 should probably be different (e.g. x0['To'] should be 16.85 instead of 290, along with the two other temperatures). Because the differential equations are only impacted by differences in temperatures, using Celsius as opposed to Kelvin shouldn't have an impact.

src/prog_models/models/battery_electrochem.py
teubert: Good catch! Fixed

src/prog_models/models/centrifugal_pump.py
```diff
@@ -121,79 +122,92 @@ class CentrifugalPump(prognostics_model.PrognosticsModel):
...
             'A': 12.7084,
             'rThrust': 1.4e-6,
             'rRadial': 1.8e-6,
-            'wA': 0.0,
-            'wThrust': 0,
-            'wRadial': 0,
         }
     }
 
+    state_limits = {
```
casonwight, line 143: For us, the parameters from the pump that exceed reasonable limits were rThrust and rRadial. Specifically, state estimators occasionally estimated negative values, which led to negative friction coefficients. From our understanding, this leads to more pump rotation meaning less pump friction, which seems incorrect.

casonwight (commented):
Our team reviewed the changes in this version. We made some comments on relevant files. Overall, we found the changes to be useful and have already implemented them from the dev branch. Things we most appreciated:

- Incorporating `state_limits`
- The callback system for parameter updates
- Wear rates as parameters for pump model
- The lazy evaluation in the simulation
- General cleanup

src/prog_models/models/centrifugal_pump.py
casonwight: Reviewed

teubert:
> Our team reviewed the changes in this version. We made some comments on relevant files. Overall, we found the changes to be useful and have already implemented them from the dev branch. Things we most appreciated:
> 
> * Incorporating `state_limits`
> * The callback system for parameter updates
> * Wear rates as parameters for pump model
> * The lazy evaluation in the simulation
> * General cleanup

@casonwight  Thank you and your team so much for your help with this review. I'm glad that you found the changes useful. 

We're going to try and resolve the last couple issues and complete our internal review process early next week. Still on our list are the following items relevant to your interests:
* The discrepancies between the pump model and the paper (that you identified). As I mention above it's possible it's the paper that's wrong. We're working to get to the bottom of it. 
* Finish specifying state bounds for the existing models

Thank you again! We really appreciate your help!

src/prog_models/models/centrifugal_pump.py
teubert: I agree- we updated the state limits to include both rThrust and rRadial

PR #122, opened 2021-08-13
State Bounds Example

wlamberson1:


examples/state_limits.py
```diff
@@ -0,0 +1,62 @@
+# Copyright © 2021 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration.  All Rights Reserved.
+
+"""
+Example demonstrating ways to use state limits. Run using the command `python -m examples.state_limits_example`
```
teubert, line 5:
```suggestion
Example demonstrating ways to use state limits. Run using the command `python -m examples.state_limits`
```

teubert (changes requested):
Great example! Good work- just one minor fix.

teubert (approved):

PR #125, opened 2021-08-13
Fix States

teubert:
* Add description for states
* Add additional state limits


teubert:
@chetankul : Could you take a look at this and see 1. is the description I added accurate, 2. Fill in the description for the missing 2 circuit states, 3. Confirm the limits I set for the states, and 4. See if I missed any limits

src/prog_models/models/battery_circuit.py
```diff
@@ -20,7 +20,10 @@ class BatteryCircuit(prognostics_model.PrognosticsModel):
         i: Current draw on the battery
 
     States: (4)
-        tb, qb, qcp, qcs
+        | tb : Battery Temperature (°C)
+        | qb : Battery Charge
+        | qcp :
```
chetankul, line 25: qb : charge stored in Capacitor Cb of the equivalent circuit model
qcp : charge stored in Capacitor Ccp of the equivalent circuit model
qcs : charge stored in Capacitor Ccs of the equivalent circuit model

chetankul (approved):

chetankul (changes requested):
qb : charge stored in Capacitor Cb of the equivalent circuit model
qcp : charge stored in Capacitor Ccp of the equivalent circuit model
qcs : charge stored in Capacitor Ccs of the equivalent circuit model

chetankul (changes requested):
since thermal runaways starts above 130C could we update the limits for tb to (-273.15, 423.15).
Rest of the limits and values look good to me

PR #131, opened 2021-09-01
Added Dynamic step size feature

teubert:
Allows user to specify a function f(t, x) -> dt that specifies the next step.

This enables use case where step sizes are dynamic (i.e., a function of time or state). For example: step size my decrease as you approach an event state

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -828,6 +828,7 @@ class PrognosticsModel(ABC):
                 inputs.append(u)
                 states.append(deepcopy(x))  # Avoid optimization where x is not copied
 
+        # configuring next_time function to define prediction time step, default is constant dt
         if callable(config['dt']):
             next_time = config['dt']
         else:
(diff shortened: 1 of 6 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -828,11 +827,19 @@ def update_all():
                 times.append(t)
                 inputs.append(u)
                 states.append(deepcopy(x))  # Avoid optimization where x is not copied
+
+        if callable(config['dt']):
```
kjjarvis, line 831: Is there a short comment that would be useful to add before this "if" statement? Something like, "configuring next_time function to define prediction time step, default is constant"

kjjarvis (approved):
This is a great new addition. The implementation in prognostics_models is straightforward and clear, and dynamic_step_size provides a detailed and easy-to-follow example.

GGorospe (approved):
New example with dynamic step size looks great!

PR #134, opened 2021-09-13
Moved limit checking into function

teubert:
v1.1 added the state bounding feature. There is some user interest in creating a separate function to apply bounds (. apply_limits(state) -> state) that is called from within __next_state(). This would enable other methods that externally agitate state (e.g., particle filters) to ensure they remain within bounds. 

## Definition of Done
- [ ] Implement . apply_bounds function in PrognosticsModel
- [ ] Update __next_state to use .apply_limits
- [ ] Create test
- [ ] Add to tutorial, example

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -404,6 +404,30 @@ class PrognosticsModel(ABC):
         dx = self.dx(x, u)
         return {key: x[key] + dx[key]*dt for key in dx.keys()}
 
+    def apply_limits(self, x):
+        """
+        Apply state bound limits. Any state outside of limits will be set to the closest limit.
+
+        Parameters
+        ----------
+        x : dict
+            state, with keys defined by model.states \n
+            e.g., x = {'abc': 332.1, 'def': 221.003} given states = ['abc', 'def']
+
+        Returns
+        -------
+        x : dict
+            Bounded state, with keys defined by model.states
+            e.g., x = {'abc': 332.1, 'def': 221.003} given states = ['abc', 'def']
+        """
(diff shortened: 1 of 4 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -404,6 +404,30 @@ def next_state(self, x, u, dt) -> dict:
...
+            state, with keys defined by model.states \n
+            e.g., x = {'abc': 332.1, 'def': 221.003} given states = ['abc', 'def']
+
+        Returns
+        -------
+        x : dict
+            Bounded state, with keys defined by model.states
+            e.g., x = {'abc': 332.1, 'def': 221.003} given states = ['abc', 'def']
+        """
+        for (key, limit) in self.state_limits.items():
```
adam-sweet, line 423: This is a user-defined function, right? If the key doesn't exist in the state, does the user get the Python "KeyError"? is that ok for your users?

teubert: A key error should be clear enough. Also, since this is in the innermost loop, I don't want to do my own check and return my own custom error message. 

In the future there might be a way for me to check each time state_limits is updated. I'll open an issue to look at that.

adam-sweet (approved):

src/prog_models/prognostics_model.py
teubert: Oh, I just realized I missed a question:

> This is a user-defined function, right?
state_limits is a user-defined python dictionary, and items() is a function available in any dictionary. 

To be properly 'pythonic' I don't check that it's a dictionary. This allows users to create their own class to replace it as desired (as long as it's utterable and has the function items() that returns keys and limits.

PR #139, opened 2021-09-26
Decouple Thrown Object from Examples

### LINKED ISSUE #137: Make Examples runnable as scripts
Problems: Some examples cannot be run alone as scripts because they're dependent on other examples. Since it's setup as a package imports have to be `from .[file_name] import [thing]` in order to support `python -m examples.[file_name]`. However, this breaks when you run a file by itself from within examples directory (e.g., press run in IDE). For that to work the command `from [file_name] import [thing]` is needed (without the period).

Multiple users have had issues with this, so I propose we remove the intra-package dependency by adding thrown_object to prog_models.models. Set it up to not be imported by default.

teubert:
Problem: Some examples cannot be run alone as scripts because they're dependent on other examples. Since it's setup as a package imports have to be `from .[file_name] import [thing]` in order to support `python -m examples.[file_name]`. However, this breaks when you run a file by itself from within examples directory (e.g., press run in IDE). For that to work the command `from [file_name] import [thing]` is needed (without the period). Multiple users have had issues with this

Solution: remove the intra-package dependency by adding thrown_object to prog_models.models. Set it up to not be imported by default (not part of __init__.__all__), and update the examples to use this version instead of the version defined in examples.new_model

Closes #137

### DIFF
--- a/examples/derived_params.py
+++ b/examples/derived_params.py
@@ -5,7 +5,7 @@
 Example demonstrating ways to use the derived parameters feature for model building. 
 """
 
-from .new_model import ThrownObject
+from prog_models.models.thrown_object import ThrownObject
 
 def run_example():
     # For this example we will use the ThrownObject model from the new_model example.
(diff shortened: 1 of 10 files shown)


kjjarvis (approved):
This change looks great - good adjustment to allow easy implementation for many users

PR #144, opened 2021-10-18
Added events example

teubert:


examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+Example further illustrating the concept of 'events' which generalizes the idea of EOL. 
+
+'Events' is the term used to describe something to be predicted. 
+Generally in the PHM community these are referred to as End of Life (EOL). 
+However, they can be much more.
+
+In the implementation from the prog_models package, events can be anything that needs to be predicted. 
+Events can represent end of life (EOL), end of mission (EOM), warning thresholds, or any event of interest (EOI). 
+
+This example shows how a events can be used in your applications.
```
mollocon44, line 14: typo - "a events"

examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+
+This example shows how a events can be used in your applications. 
+"""
+from prog_models.models import BatteryElectroChemEOD
+
+def run_example():
+    # Example: Warning thresholds
+    # In this example we will use the battery model
+    # We of course are interested in end of discharge, but for this example we
+    # have a requirement that says the battery must not fall below 5% SOC
```
mollocon44, line 22: I don't think you've explained SOC==State of Charge yet

examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+    # Additionally, we have two warning thresholds (yellow and red)
+
+    YELLOW_THRESH = 0.15
```
mollocon44, line 25: This would be a great spot to remind/explain that the battery SOC goes from 0 to 1 - I think I was confused by that at first

examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Step 1: Extend the battery model to define the additional events
+    class MyBatt(BatteryElectroChemEOD):
+        events = BatteryElectroChemEOD.events + ['EOD_warn_yellow', 'EOD_warn_red']
+
+        def event_state(self, state):
+            # Get event state from parent
+            event_state = super().event_state(state)
+
+            # Add yellow and red states by scaling EOD state
```
mollocon44, line 36: This is a little hard to follow - is it scaling to make the EOD effectively 0% SOC?

examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+            # Add yell and red states from event_state
+            event_state = self.event_state(x)
+            t_met['EOD_warn_yellow'] = event_state['EOD_warn_yellow'] <= 0
+            t_met['EOD_warn_red'] = event_state['EOD_warn_red'] <= 0
+
+            return t_met
+
+    # Step 2: Use it
+    # 2a: Setup model
+    def future_loading(t, x=None):
```
mollocon44, line 56: Might want to remind user that this is future current loading

examples/events.py
```diff
@@ -0,0 +1,83 @@
...
+            i = 1
+        elif (t < 1800):
+            i = 4
+        elif (t < 3000):
+            i = 2     
+        else:
+            i = 3
+        return {'i': i}
+    
+    first_output = {'t': 18.95, 'v': 4.183}
```
mollocon44, line 70: Is first_output the initial state of the battery?

mollocon44 (approved):
A really nice example overall - most of my comments are on leaving extra comments, with just two spots where I was a little confused  about what the code was doing (37/38 and 70)

examples/events.py
mollocon44: Ohhhh it's scaling between voltage levels and percentages, isn't it. This could probably use an extra comment, I got a bit lost here.

examples/events.py
teubert: It's the first output (i.e., measurement), used to estimate first state. It's not strictly needed here and doesn't add anything to the example, so I'm going to remove it.

examples/events.py
```diff
@@ -67,11 +76,10 @@ def future_loading(t, x=None):
             i = 3
         return {'i': i}
     
-    first_output = {'t': 18.95, 'v': 4.183}
     m = MyBatt()
 
     # 2b: Simulate to threshold
-    (times, inputs, states, outputs, event_states) = m.simulate_to_threshold(future_loading, first_output, threshold_keys=['EOD'], print = True)
+    (_, _, _, _, event_states) = m.simulate_to_threshold(future_loading, threshold_keys=['EOD'], print = True)
```
mollocon44, line 82: For clarity, it might be nice to leave (times, inputs, states, outputs, event_states) as a reminder of what the output is, unless the code needs it to be shorter

teubert:
```suggestion
    (times, inputs, states, outputs, event_states) = m.simulate_to_threshold(future_loading, threshold_keys=['EOD'], print = True)
```

PR #150, opened 2021-10-22
extend sim result

teubert:
Extend SimResult class to include some of the functions present in list, explicitly throwing exception when it's not implemented.

Also, added support for empty simresult

Note: Pop and extend are needed for the rework of MonteCarlo


GGorospe (approved):
Recommend adding more context to the following inline comment:
"Get the index of the first sample where other occurs."

PR #157, opened 2021-11-05
Make initial output and input optional in initialize

teubert:
Make initial output and input optional for some models, propagate changes to simulate_to and examples. 

For models that dont use output to estimate initial state, we shouldn't have to supply it. 

This enables using default first state in simulation. A feature requested in PaaS

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -645,7 +645,7 @@ class PrognosticsModel(ABC):
         future_loading_eqn : callable
             Function of (t) -> z used to predict future loading (output) at a given time (t)
         first_output : dict, optional
-            First measured output, needed to initialize state for some classes. Can be ommited for classes that dont use this
+            First measured output, needed to initialize state for some classes. Can be omitted for classes that dont use this
         options: kwargs, optional
             Configuration options for the simulation \n
             Note: configuration of the model is set through model.parameters \n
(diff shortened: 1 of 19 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -644,6 +644,8 @@ def simulate_to(self, time, future_loading_eqn, first_output, **kwargs) -> tuple
             e.g., time = 200
         future_loading_eqn : callable
             Function of (t) -> z used to predict future loading (output) at a given time (t)
+        first_output : dict, optional
+            First measured output, needed to initialize state for some classes. Can be ommited for classes that dont use this
```
kjjarvis, line 648: Very minor but I think "omitted" only has one "m"

src/prog_models/prognostics_model.py
```diff
@@ -695,16 +697,16 @@ def simulate_to(self, time, future_loading_eqn, first_output, **kwargs) -> tuple
...
         Simulate prognostics model until any or specified threshold(s) have been met
 
         Parameters
         ----------
         future_loading_eqn : callable
             Function of (t) -> z used to predict future loading (output) at a given time (t)
-        first_output : dict
-            First measured output, needed to initialize state
+        first_output : dict, optional
+            First measured output, needed to initialize state for some classes. Can be ommited for classes that dont use this
```
kjjarvis, line 709: "omitted"

kjjarvis (approved):

src/prog_models/prognostics_model.py
teubert:
```suggestion
            First measured output, needed to initialize state for some classes. Can be omitted for classes that dont use this
```

src/prog_models/prognostics_model.py
teubert:
```suggestion
            First measured output, needed to initialize state for some classes. Can be omitted for classes that dont use this
```

PR #172, opened 2021-11-30
prog_models v1.2.1: Added support to python3.10 

teubert:
* Added support to python3.10 
* Added missing example to tests
* Update version #
* Put version #s in quotes for clarity in .github workflow (otherwise only 3.10 would be in quotes)


.github/workflows/python-package.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
```
jason-watkins, line 51: Yaml is confusing enough as it is, but I think the quotes aren't necessary here. If they are for some reason, I think quoting every version should work and would remove a potential source of confusion.

jason-watkins (approved):

.github/workflows/python-package.yml
teubert: The quotes are unfortunately necessary. Otherwise this resolves to python v3.1 instead of 3.10. I'm fine with quotes for every version.

teubert: Done- I put the version #s in quotes for clarity.

PR #179, opened 2021-12-23
Implement Linear Model

teubert:
Implement a linear model abstract base class for models that can be described as a linear timeseries. Created an example demonstrating this

### DIFF
--- a/src/prog_models/linear_model.py
+++ b/src/prog_models/linear_model.py
@@ -9,19 +9,19 @@ import numpy as np
 class LinearModel(PrognosticsModel, ABC):
     """
     A linear prognostics model. Used when behavior can be described using a simple linear time-series model defined by the following equations:
-        * dx/dt = Ax + Bu + C
-        * z = Dx + E
+        * dx/dt = Ax + Bu + E
+        * z = Cx + D
         * es = Fx + G
     where x is state, u is input, z is output and es is event state
 
     Linear Models must inherit from this class and define the following properties:
-        * A: 2-d numpy.array[float] 
-        * B: 2-d numpy.array[float], optional (zeros by default)
-        * C: 2-d numpy.array[float], optional (zeros by default)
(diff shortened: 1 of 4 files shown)


teubert:
Note: Should fail on released version of prog_algs, but pass on dev

src/prog_models/linear_model.py
```diff
@@ -0,0 +1,84 @@
...
+
+from . import PrognosticsModel
+from abc import ABC, abstractmethod
+import numpy as np
+
+
+class LinearModel(PrognosticsModel, ABC):
+    """
+    A linear prognostics model. Used when behavior can be described using a simple linear time-series model defined by the following equations:
+        * dx/dt = Ax + Bu + C
```
matteocorbetta, line 12: Notation: typical notation is sligthly different. I'd recommend following the standard one, and I can rewrite LinearModel accordingly.

dx/dt = Ax + Bu + E
y = Cx + D
es = Fx + G

src/prog_models/linear_model.py
```diff
@@ -0,0 +1,84 @@
...
+        * z = Dx + E
+        * es = Fx + G
+    where x is state, u is input, z is output and es is event state
+
+    Linear Models must inherit from this class and define the following properties:
+        * A: 2-d numpy.array[float]
```
matteocorbetta, line 18: We could add matrix dimensions; makes life easier for the user

A -> n_states x n_states
B -> n_states x n_inputs
E -> n_states x n_state_disturbances 

C -> n_outputs x n_states
D -> n_outputs x n_meas_disturbances

F -> n_eventstates x n_states
G -> n_eventstates x n_es_disturbances

matteocorbetta (changes requested):
Requesting changing on matrix notations (files edited) before merging.

matteocorbetta (approved):

PR #183, opened 2022-01-12
Add warning when setting param that will be overwritten

teubert:
Some parameters are overwritten in model extensions, adding warning that notifies user if they're setting a parameter for which this is true.

Also, removed 2 parameters that were unnecessarily being set in output method


teubert:
Failed tests are ok, related to open issue in prog_algs where particle filter occasionally is outside of test range.

kjjarvis (approved):

PR #190, opened 2022-01-29
Linear Model: Matrix size check

lawrence-hwang:
PR for adding Matrix dimension checking for Linear Model property(s) A,B,C,D,E,F (also supports None check for F)


src/prog_models/linear_model.py
```diff
@@ -28,6 +29,53 @@ class LinearModel(PrognosticsModel, ABC):
...
+    def _propertyCheck(self, matrix, rowsCount, colsCount, notes):
+        """
+        matrix: Input matrix to check dimensions of (e.g. self.A, self.B, etc)
+        rowsCount: Row count to check matrix against
+        colsCount: Column count to check matrix against
+        notes: List of strings containing information for exception message debugging
+        """
+        raiseInformative = false
+        # first perform col check for each row
+        for subArr in matrix:
```
teubert, line 72: I think you can use matrix.ndim and .shape for this check. That will probably be a little cleaner.

src/prog_models/linear_model.py
```diff
@@ -28,6 +29,53 @@ class LinearModel(PrognosticsModel, ABC):
...
+            if len(subArr) != colsCount:
+                raiseInformative = true # we could also track which arr specifically if needed, put that in error message with %s
+                break
+        if (raiseInformative or len(matrix) != rowsCount): # check along cols, rows
+            raise Exception("Matrix size check failed: @property {} dimensions improperly formed along {} x {}.".format(notes[0],notes[1],notes[2]))
```
teubert, line 77: This is great, very informative exception message

src/prog_models/linear_model.py
```diff
@@ -28,6 +29,53 @@ class LinearModel(PrognosticsModel, ABC):
...
     """
 
+    def __init__(self):
+        super().__init__()
+
+        """
+        Member matrices size checks for user specified @property(s)
+        """
+        # @property A Matrix Check
+        if (self.A.any()):
```
teubert, line 39: What role does the any play here?

src/prog_models/linear_model.py
```diff
@@ -28,6 +29,53 @@ class LinearModel(PrognosticsModel, ABC):
         * events:  list[str] - event keys
     """
 
+    def __init__(self):
+        super().__init__()
+
+        """
```
teubert, line 35: Use a regular comment here instead of docstring. Docstrings are used as labels at the beginning of a file/class/function

src/prog_models/linear_model.py
```diff
@@ -1,6 +1,7 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the
 # National Aeronautics and Space Administration.  All Rights Reserved.
 
+from sympy import false, true
```
teubert, line 4: What role does this play? I've never seen sympy true and false variables used. Will the language-included `True` and `False` statements work for your use case?

tests/test_base_models.py
```diff
@@ -3,6 +3,7 @@
 import unittest
 
 from attr import Attribute
+from sympy import true
```
teubert, line 6: I think VSCode added this again

tests/test_base_models.py
```diff
@@ -915,19 +916,159 @@ def event_state(self, x):
...
+        with self.assertRaises(TypeError):
+            m.E = true # boolean
+            m.matrixCheck()
+        # @F
+        with self.assertRaises(TypeError):
+            m.F = "[[0, 1], [0, 0]]" # string
+            m.matrixCheck() 
         with self.assertRaises(TypeError):
-            m.A = [[0, 1], [0, 0]] # standard 2-D array
+            m.F = None # None
```
teubert, line 1004: F can be none (if users specify an event_state function). This shouldn't raise a TypeError

tests/test_base_models.py
```diff
@@ -1,6 +1,9 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
 import unittest
+
+from attr import Attribute
```
teubert, line 5: This adds a new dependency to prog_models. Is there a way to do this without adding a new dependency. Does the python built in setattr() function serve the same purpose?

lawrence-hwang: I think this was erroneously added too while I was testing - removed.

teubert (approved):
Looks good!

PR #193, opened 2022-02-02
Cache array sizes in prognostics_model and all children class models

lawrence-hwang:
Member variables for caching array sizes of events, inputs, outputs, state, and performance.

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -33,7 +33,7 @@ class PrognosticsModelParameters(UserDict):
             self[key] = deepcopy(value)
 
         # Add class variables for length of inputs, outputs, events, state, and performance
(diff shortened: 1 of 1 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -32,6 +32,9 @@ def __init__(self, model, dict_in = {}, callbacks = {}):
             # Deepcopy is needed here to force copying when value is an object (e.g., dict)
             self[key] = deepcopy(value)
 
+        # Add class variables for length of inputs, outputs, events, state, and performance
+        self.lenStates = len(self.states)
```
teubert, line 36:
```suggestion
        self.n_states = len(self.states)
```
Just a preference thing- I prefer the notation n_ instead of len...

Also note that python naming conventions use underscores for variable names.

PR #194, opened 2022-02-03
Caching array sizes as class variables in PrognosticsModel

lawrence-hwang:
Caching sizes of each significant array in PrognosticsModel and all children classes, refining how lengths of each array are obtained throughout the package.

### DIFF
--- a/src/prog_models/linear_model.py
+++ b/src/prog_models/linear_model.py
@@ -78,7 +78,8 @@ class LinearModel(PrognosticsModel, ABC):
 
     @property
     def E(self):
-        return np.zeros((self.n_states, 1))
+        n_states = self.n_states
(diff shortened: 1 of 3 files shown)


src/prog_models/linear_model.py
```diff
@@ -73,13 +73,13 @@ def A(self):
 
     @property
     def B(self):
-        n_inputs = len(self.inputs)
-        n_states = len(self.states)
+        n_inputs = self.n_inputs
```
teubert, line 76: I don't think the declaration adds anything anymore. Let's combine it into line 78, so 78 becomes:
```python
return np.zeros((self.n_states, self.n_inputs))
```

Let's make the same change for the other ones too.

teubert (approved):

PR #196, opened 2022-02-03
Debugging __init__() of LinearModel introduced in matrix size checking fork

lawrence-hwang:


teubert (approved):

teubert:
prog_algs v1.2 is incomparable with prog_models v1.3. The prog_algs release tests will fail. This is expected.

PR #198, opened 2022-02-08
Extending tests for SimResult

lawrence-hwang:
Testing SimResult's extend public method

### DIFF
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -71,17 +71,8 @@ class SimResult(UserList):
         Args:
             target (int): Value of element to be removed.
         """
+        self.data.pop(self.times.index(target))
         self.times.remove(target)
-        self.data.remove(target)
-
-    # def remove(self, index : int) -> None:
-    #     """Remove an element
-
-    #     Args:
-    #         index (int): Index of element to be removed.
-    #     """
-    #     self.times.remove(index)
-    #     self.data.remove(index)
 
     def clear(self) -> None:
         """Clear the SimResult"""
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -156,7 +156,7 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
         self.__data = None
         self.states.extend(deepcopy(other.states))  # lgtm [py/modification-of-default-value]
 
-    def pop(self, index = -1):
+    def pop(self, index : int = -1):
         """Remove an element. If data hasn't been cached, remove the state - so it wont be calculated
 
         Args:
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -71,8 +71,8 @@ class SimResult(UserList):
         Args:
             target (int): Value of element to be removed.
         """
-        self.data.pop(self.times.index(target))
-        self.times.remove(target)
+        self.times.pop(self.data.index(target))
+        self.data.remove(target)
 
     def clear(self) -> None:
         """Clear the SimResult"""
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -173,7 +173,6 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
         self.states.pop(target_index)
         if self.__data is not None:
             self.__data.remove(target)
-        # self.data.remove(target)
 
     @property
     def data(self):
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -69,7 +69,7 @@ class SimResult(UserList):
         """Remove an element
 
         Args:
-            target (int): Value of element to be removed.
+            target: Value of element to be removed.
         """
         self.times.pop(self.data.index(target))
         self.data.remove(target)
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -65,14 +65,23 @@ class SimResult(UserList):
         self.times.pop(index)
         return self.data.pop(index)
     
-    def remove(self, index : int) -> None:
+    def remove(self, target : int) -> None:
         """Remove an element
 
         Args:
-            index (int): Index of element to be removed.
+            target (int): Value of element to be removed.
         """
-        self.times.remove(index)
-        self.data.remove(index)
+        self.times.remove(target)
+        self.data.remove(target)
+
+    # def remove(self, index : int) -> None:
+    #     """Remove an element
+
+    #     Args:
+    #         index (int): Index of element to be removed.
+    #     """
+    #     self.times.remove(index)
+    #     self.data.remove(index)
 
     def clear(self) -> None:
         """Clear the SimResult"""
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -166,7 +166,7 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
         """Remove an element
 
         Args:
-            target (int): Value of element to be removed.
+            target: Value of element to be removed.
         """
         target_index = self.data.index(target)
         self.times.pop(target_index)
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -162,7 +162,7 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
             return self.__data.pop(index)
         return self.fcn(x)
 
-    def remove(self, target : int) -> None:
+    def remove(self, target) -> None:
         """Remove an element
 
         Args:
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -65,7 +65,7 @@ class SimResult(UserList):
         self.times.pop(index)
         return self.data.pop(index)
     
-    def remove(self, target : int) -> None:
+    def remove(self, target) -> None:
         """Remove an element
 
         Args:
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -162,6 +162,19 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
             return self.__data.pop(index)
         return self.fcn(x)
 
+    def remove(self, target : int) -> None:
+        """Remove an element
+
+        Args:
+            target (int): Value of element to be removed.
+        """
+        target_index = self.data.index(target)
+        self.times.pop(target_index)
+        self.states.pop(target_index)
+        if self.__data is not None:
+            self.__data.remove(target)
+        # self.data.remove(target)
+
     @property
     def data(self):
         """
--- a/tests/test_sim_result.py
+++ b/tests/test_sim_result.py
@@ -76,6 +76,10 @@ class TestSimResult(unittest.TestCase):
         self.assertEqual(result.data, [0.0, 2.5, 7.5, 10.0])
         result.pop() # Test default index -1 (last element)
         self.assertEqual(result.data, [0.0, 2.5, 7.5])
+        result.pop(-1) # Test argument of index -1 (last element)
+        self.assertEqual(result.data, [0.0, 2.5])
+        result.pop(0) # Test argument of 0
+        self.assertEqual(result.data, [2.5])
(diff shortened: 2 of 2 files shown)


src/prog_models/sim_result.py
```diff
@@ -65,14 +65,23 @@ def pop(self, index : int = -1) -> dict:
...
         Args:
-            index (int): Index of element to be removed.
+            target (int): Value of element to be removed.
         """
-        self.times.remove(index)
-        self.data.remove(index)
+        self.times.remove(target)
+        self.data.remove(target)
+
+    # def remove(self, index : int) -> None:
```
teubert, line 77: Make sure you remove this commented section

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+    def test_index(self):
+        NUM_ELEMENTS = 5 # Creating two result objects
+        time = list(range(NUM_ELEMENTS))
+        state = [i * 2.5 for i in range(NUM_ELEMENTS)]
+        result = SimResult(time, state)
+        # other_arg = {2.5:1}
+        # other_arg = {1:2.5}
+        other_arg = 2.5
+
+        self.assertEqual(result.index(other_arg), 1)
```
teubert, line 67: Let's test the behavior for when the other arg doesn't exist as well (exception should be raised)

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+        result.pop() # Test default index -1 (last element)
+        self.assertEqual(result.data, [0.0, 2.5, 7.5])
+        self.assertRaises(IndexError, result.pop, 5) # Test specifying an invalid index value
+        self.assertRaises(IndexError, result.pop, 3)
+        self.assertRaises(TypeError, result.pop, "5") # Test specifying an invalid index type
+        self.assertRaises(TypeError, result.pop, [0,1])
+        self.assertRaises(TypeError, result.pop, {})
+        self.assertRaises(TypeError, result.pop, set())
+        self.assertRaises(TypeError, result.pop, 1.5)
+        # self.assertRaises(TypeError, result.pop, True) # This removes index 1 because boolean True, is this okay?
```
teubert, line 86:
```suggestion
```

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+        state = [i * 2.5 for i in range(NUM_ELEMENTS)]
+        result = SimResult(time, state)
+
+        result.pop(2) # Test specified index
+        self.assertEqual(result.data, [0.0, 2.5, 7.5, 10.0])
+        result.pop() # Test default index -1 (last element)
+        self.assertEqual(result.data, [0.0, 2.5, 7.5])
+        self.assertRaises(IndexError, result.pop, 5) # Test specifying an invalid index value
+        self.assertRaises(IndexError, result.pop, 3)
+        self.assertRaises(TypeError, result.pop, "5") # Test specifying an invalid index type
```
teubert, line 81: I would also add a negative index to this (should count from back)

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+        self.assertEqual(result.time(3), result.times[3])
+        self.assertEqual(result.time(4), result.times[4])
+
+        self.assertRaises(TypeError, result.time, ) # Test no input given
+        self.assertRaises(TypeError, result.time, "0") # Tests specifying an invalid index type 
+        self.assertRaises(TypeError, result.time, [0,1])
+        self.assertRaises(TypeError, result.time, {})
+        self.assertRaises(TypeError, result.time, set())
+        self.assertRaises(TypeError, result.time, 1.5)
+        # self.assertRaises(TypeError, result.time, True) # Test no input given
```
teubert, line 146:
```suggestion
```

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+
+    def test_plot(self):
+        NUM_ELEMENTS = 5 # Creating two result objects
+        time = list(range(NUM_ELEMENTS))
+        state = [i * 2.5 for i in range(NUM_ELEMENTS)]
+        result = SimResult(time, state)
+        # INCOMPLETE
```
teubert, line 153: As we discussed, minimum test is just running the plot function and making sure it doesn't raise an exception. Something more would be nice, but not essential

src/prog_models/sim_result.py
```diff
@@ -65,14 +65,23 @@ def pop(self, index : int = -1) -> dict:
         self.times.pop(index)
         return self.data.pop(index)
     
-    def remove(self, index : int) -> None:
+    def remove(self, target : int) -> None:
         """Remove an element
 
         Args:
-            index (int): Index of element to be removed.
+            target (int): Value of element to be removed.
```
teubert, line 72: Great catch!

tests/test_sim_result.py
```diff
@@ -36,6 +36,121 @@ def test_pickle(self):
...
+        # self.assertRaises(TypeError, result.remove, "5") # Tests specifying an invalid index type
+        # self.assertRaises(TypeError, result.remove, [0,1])
+        # self.assertRaises(TypeError, result.remove, {})
+        # self.assertRaises(TypeError, result.remove, set())
+        # self.assertRaises(TypeError, result.remove, 1.5)
+
+        # self.assertRaises(IndexError, result.remove, 5) # Test specifying an invalid value
+        # self.assertRaises(IndexError, result.remove, 3)
+        # self.assertRaises(TypeError, result.remove, True) # This removes index 1 because boolean True, is this okay?
+
```
teubert, line 115: I also recommend you check the not implemented functions. Just make sure that they raise the expected exception

teubert (commented):
It's looking great. Besides finishing the remove test (which I know you're working on), I have just a few suggestions (some of which we discussed today).

tests/test_sim_result.py
```diff
@@ -1,6 +1,8 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
 import unittest
+
+from attr import Attribute
```
teubert, line 5:
```suggestion
```
I think your IDE added this

tests/test_sim_result.py
```diff
@@ -36,7 +38,245 @@ def test_pickle(self):
...
+
+    # Tests for LazySimResult
+    def test_lazy_data_fcn(self):
+        def f(x):
+            return x * 2
+        NUM_ELEMENTS = 5
+        time = list(range(NUM_ELEMENTS))
+        state = [i * 2.5 for i in range(NUM_ELEMENTS)]
+        result = LazySimResult(f, time, state)
+        self.assertEqual(result.data, [0.0, 5.0, 10.0, 15.0, 20.0])
```
teubert, line 202: Can you check is_cached before and after this line. It should be false beforehand and true afterwards

tests/test_sim_result.py
```diff
@@ -36,7 +38,245 @@ def test_pickle(self):
...
+        self.assertEqual(result.states, [0.0, 2.5, 5.0, 7.5, 10.0])
+        self.assertEqual(result2.times, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        self.assertEqual(result2.data, [0, 25, 50, 75, 100, 125, 150, 175, 200, 225])
+        self.assertEqual(result2.states, [0, 5, 10, 15, 20, 25, 30, 35, 40, 45])
+
+        result.extend(result2)
+        self.assertEqual(result.times, [0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9]) # Assert data is correct after extending
+        self.assertEqual(result.data, [0.0, 5.0, 10.0, 15.0, 20.0, 0, 10, 20, 30, 40, 50, 60, 70, 80, 90])
+        self.assertEqual(result.states, [0.0, 2.5, 5.0, 7.5, 10.0, 0, 5, 10, 15, 20, 25, 30, 35, 40, 45])
+
```
teubert, line 246: Another test to add. Extending should reset cache. 

So, if you do this:
   * Build lazy sim result (line 234)
   * Access data (lines 235-241)
   * Check is_cached (should be true here)
   * Extend (line 242)
   * Check is_cached (should be false here)
   * Access data (lines 243-245)

tests/test_sim_result.py
```diff
@@ -36,7 +38,245 @@ def test_pickle(self):
...
+        self.assertEqual(result.states, [5.0])
+        # Test erroneous input
+        self.assertRaises(IndexError, result.pop, 5) # Test specifying an invalid index value
+        self.assertRaises(IndexError, result.pop, 3)
+        self.assertRaises(TypeError, result.pop, "5") # Test specifying an invalid index type
+        self.assertRaises(TypeError, result.pop, [0,1])
+        self.assertRaises(TypeError, result.pop, {})
+        self.assertRaises(TypeError, result.pop, set())
+        self.assertRaises(TypeError, result.pop, 1.5)
+
```
teubert, line 279: Same test as above, but with pop. Make sure pop resets the cache

tests/test_sim_result.py
```diff
@@ -73,6 +313,47 @@ def f(x):
...
+        self.assertEqual(result.times, [2, 3])
+        self.assertEqual(result.data, [10.0, 15.0])
+
+        self.assertRaises(TypeError, result.remove, ) # Test no index specified
+        self.assertRaises(ValueError, result.remove, 20.0) # Test nonexistent data value
+        self.assertRaises(ValueError, result.remove, -1) # Type checking negated as index searches for element in list
+        self.assertRaises(ValueError, result.remove, "5") # Thus all value types allowed to be searched
+        self.assertRaises(ValueError, result.remove, [0,1])
+        self.assertRaises(ValueError, result.remove, {})
+        self.assertRaises(ValueError, result.remove, set())
```
teubert, line 341: Same test as with pop

teubert (changes requested):
Just about there. Just one bug that's causing the tests to fail (the extra import added by your IDE) and a few additional suggested LazySimResult tests

src/prog_models/sim_result.py
```diff
@@ -65,14 +65,14 @@ def pop(self, index : int = -1) -> dict:
         self.times.pop(index)
         return self.data.pop(index)
     
-    def remove(self, index : int) -> None:
+    def remove(self, target : int) -> None:
```
teubert, line 68:
```suggestion
    def remove(self, target) -> None:
```
I just realized, for remove the target should actually be type dict (since it's removing by value. I think it's best to keep the hint off for now

src/prog_models/sim_result.py
```diff
@@ -65,14 +65,14 @@ def pop(self, index : int = -1) -> dict:
...
         """Remove an element
 
         Args:
-            index (int): Index of element to be removed.
+            target (int): Value of element to be removed.
```
teubert, line 72:
```suggestion
            target: Value of element to be removed.
```
See above

src/prog_models/sim_result.py
```diff
@@ -162,6 +162,19 @@ def pop(self, index = -1):
             return self.__data.pop(index)
         return self.fcn(x)
 
+    def remove(self, target : int) -> None:
```
teubert, line 165:
```suggestion
    def remove(self, target) -> None:
```
See above

src/prog_models/sim_result.py
```diff
@@ -162,6 +162,19 @@ def pop(self, index = -1):
...
+        """Remove an element
+
+        Args:
+            target (int): Value of element to be removed.
```
teubert, line 169:
```suggestion
            target: Value of element to be removed.
```
See above

src/prog_models/sim_result.py
```diff
@@ -162,6 +162,19 @@ def pop(self, index = -1):
...
+        """
+        target_index = self.data.index(target)
+        self.times.pop(target_index)
+        self.states.pop(target_index)
+        if self.__data is not None:
+            self.__data.remove(target)
+        # self.data.remove(target)
```
teubert, line 176:
```suggestion
```

teubert (changes requested):
A few small changes

PR #204, opened 2022-02-10
Enhancing remove by enabling named arguments and removal using times,states

lawrence-hwang:


src/prog_models/sim_result.py
```diff
@@ -181,21 +181,27 @@ def remove(self, d = None, t = None, s = None) -> None:
             t: Time value to be removed.
             s: State value to be removed.
         """ 
-        if (not [i for i in (d,t,s) if i is not None or None]):
+        if (not [i for i in (d,t,s) if i is not None or None]) or (not [i for i in (d,t) if i is None]) or (not [i for i in (t,s) if i is None]) or (not [i for i in (d,s) if i is None]):
```
lawrence-hwang, line 184: Perhaps there is a better way to check this?

Its checking to make sure that (d,t,s) representing the (data,times,state) targets meet these conditions:
- All three cannot be specified
- All three cannot be None
- Two out of the three are specified (this specifically - could we improve somehow? Currently using 3 list comprehensions to check all three combinations of (d,t), (d,s), (t,s))

teubert: hmmm... probably is a better way, but this can be tricky. How about:

```python
if sum([i is None for i in (d, t, s)]) != 1:
```

That would count how many variables are not none (should only be one). would that work?

src/prog_models/sim_result.py
```diff
@@ -165,17 +173,35 @@ def pop(self, index : int = -1):
...
-        target_index = self.data.index(target)
-        self.times.pop(target_index)
-        self.states.pop(target_index)
-        if self.__data is not None:
-            self.__data.remove(target)
+            d: Data value to be removed.
+            t: Time value to be removed.
+            s: State value to be removed.
+        """ 
+        if (not [i for i in (d,t,s) if i is not None or None]) or (sum([i is None for i in (d, t, s)]) != 2):
```
teubert, line 184:
```suggestion
        if sum([i is None for i in (d, t, s)]) != 2:
```
I don't think we need the first part. Just make sure there are exactly 2 Nones

src/prog_models/sim_result.py
```diff
@@ -165,17 +173,35 @@ def pop(self, index : int = -1):
...
+            target_index = self.states.index(s)
+            self.times.pop(target_index)
+            self.states.pop(target_index)
+            if self.__data is not None:
+                self.__data.pop(target_index)
+        else:
+            target_index = self.data.index(d)
+            self.times.pop(target_index)
+            self.states.pop(target_index)
+            if self.__data is not None:
```
teubert, line 203: I think the if statement on Line 203 isn't actually necessary. __data is resolved on line 200 when you access it to check the index. Instead you can pop __data without checking if its none.

teubert (changes requested):
Looks pretty good! A few minor optimizations.

src/prog_models/sim_result.py
lawrence-hwang: Tested, this is true. Making change

teubert (approved):

PR #206, opened 2022-02-11
Implementation of LazySimResult extending a SimResult object

lawrence-hwang:
Enable SimResult to be extended by LazySimResult.

If possible, extend LazySimResult with SimResult. If not, add check to prevent trying.

### DIFF
--- a/src/prog_models/sim_result.py
+++ b/src/prog_models/sim_result.py
@@ -158,15 +158,21 @@ class LazySimResult(SimResult):  # lgtm [py/missing-equals]
 
     def extend(self, other):
         """
-        Extend the LazySimResult with another LazySimResult or SimResult object
+        Extend the LazySimResult with another LazySimResult object
+        Raise ValueError if SimResult is passed
 
         Args:
-            other (SimResult/LazySimResult)
+            other (LazySimResult)
 
         """
-        self.times.extend(deepcopy(other.times))  # lgtm [py/modification-of-default-value]
-        self.__data = None
-        self.states.extend(deepcopy(other.states))  # lgtm [py/modification-of-default-value]
+        if (isinstance(other, self.__class__)):
+            self.times.extend(deepcopy(other.times))  # lgtm [py/modification-of-default-value]
+            self.__data = None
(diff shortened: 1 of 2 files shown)


src/prog_models/sim_result.py
```diff
@@ -1,5 +1,7 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 from collections import UserList
+
+from numpy import isin
```
teubert, line 4: I don't think you use this at all

src/prog_models/sim_result.py
```diff
@@ -50,8 +52,11 @@ def extend(self, other) -> None:
             other (SimResult/LazySimResult)
 
         """
-        self.times.extend(other.times)
-        self.data.extend(other.data)
+        if other.__class__ in [SimResult, LazySimResult]:
```
teubert, line 55: A good addition!

src/prog_models/sim_result.py
```diff
@@ -149,14 +154,33 @@ def is_cached(self):
...
+        Raise ValueError if SimResult is passed
+
+        Args:
+            other (LazySimResult)
+
+        """
+        if (isinstance(other, self.__class__)):
+            self.times.extend(deepcopy(other.times))  # lgtm [py/modification-of-default-value]
+            self.states.extend(deepcopy(other.states))  # lgtm [py/modification-of-default-value]
+            if self.__data is None or not other.is_cached():
```
teubert, line 176: Now that you added this feature, lets add a test for it. If you do the following:

* Create 2 Lazy Sim Results without accessing data -> extend one from the other -> check extended lazysimresult.is_cached - Should be false
* Create 2 Lazy Sim Results access 1st one  -> extend one from the other -> check extended lazysimresult.is_cached - Should be false
* Create 2 Lazy Sim Results access 2nd one  -> extend one from the other -> check extended lazysimresult.is_cached - Should be false
* * Create 2 Lazy Sim Results access both one  -> extend one from the other -> check extended lazysimresult.is_cached - Should be true

teubert (approved):

PR #208, opened 2022-02-14
Feature for simulate_to* to return namedtuples instead of basic tuples

lawrence-hwang:
Implementing change for simulate_to* methods in prog_models/prognostics_model.py to return namedtuples instead of basic tuples. Motivation for change is ease of readability and access to objects in the return from the method's result.

### DIFF
--- a/tests/test_battery.py
+++ b/tests/test_battery.py
@@ -1,6 +1,8 @@
 # Copyright © 2020 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
 import unittest
+
+from pandas import array
 from prog_models.models import BatteryCircuit, BatteryElectroChem, BatteryElectroChemEOL, BatteryElectroChemEOD, BatteryElectroChemEODEOL
 
 def future_loading(t, x=None):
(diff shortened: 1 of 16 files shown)


tests/test_battery.py
```diff
@@ -1,6 +1,8 @@
 # Copyright © 2020 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
 import unittest
+
+from pandas import array
```
teubert, line 5:
```suggestion
```

I think this was accidental.

teubert (changes requested):
Wow, this worked out really cleanly. Great work. 

Could you also look at the examples and make changes to a few or all of them to use the named tuple notation?

lawrence-hwang:
> Wow, this worked out really cleanly. Great work.
> 
> Could you also look at the examples and make changes to a few or all of them to use the named tuple notation?

namedtuple Notation applied to only a few of the examples:
    - Call simulate_to just to call, don’t unpack returned items for separate use (no changes applied):
        - benchmarking, dynamic_step_size, simulate_to_threshold, sim_battery_eol, sim_valve, sim, vectorized
    - Don’t use simulate_to at all (no changes applied):
        - derived_params, param_est
    - Examples switched to Namedtuple syntax (changes applied):
        - events,  future_loading, model_gen, new_model (unpacking names changed), noise, sensitivity, sim_pump, state_limits, visualize

teubert (approved):

Issue #207, opened 2022-02-11
LazySimResult.extend() should check argument for same function other.fcn()


lawrence-hwang:
LazySimResult can only be extended by another LazySimResult that has the same function fcn. Not enough to only check name, functionality could differ. Even if functionality is the same, implementation could differ to produce same result.

Checking ideas:
- Bytecode using dis.dis, inspect.getsource(), fcn.__code__.co_code https://stackoverflow.com/questions/20059011/check-if-two-python-functions-are-equal
- Check results? Could amount to infinite/inefficient checking

lawrence-hwang:
After attempting various methods of function checking, we have come to the conclusion that there is no robust method of verifying two functions are the same EXCEPT manually vetting all inputs through two functions to ensure equivalent outputs. Even this method is not perfect, and only raises confidence in sameness from the given inputs. 

The solution to LazySimResult.extend() fcn checking is therefore reminding the user in code comments that a LazySimResult extending another MUST have the same fcn function.

Testing was done by creating two nearly identical functions performing the same simple task differently. As demonstrated in closed PR#209 and branch 'feature/LazySimResult_other_fcn_check', the following methods fail in properly identifying matching functions:
- Byte code via Bytecode class disassembly (Bytecode.dis(), Bytecode.info, Bytecode.codeobj)
- Byte code via obj.__code__.co_code
- inspect getsource()
- dis.dis

PR #209, opened 2022-02-15
LazySimResult.extend() check for same fcn

lawrence-hwang:
Checking experimentation to ensure fcn of both original LazySimResult and other LazySimResult have the same fcn before extending can be performed.

### DIFF
--- a/tests/test_sim_result.py
+++ b/tests/test_sim_result.py
@@ -503,10 +503,20 @@ class TestSimResult(unittest.TestCase):
 
         arr = [0,1,2,3,4]
         i = 10
-        self.assertEquals(f1(i,arr), f2(i,arr))
+        self.assertEqual(f1(i,arr), f2(i,arr)) # produce same result
 
         # comparing references to same function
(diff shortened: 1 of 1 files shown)


lawrence-hwang:
Issue #207 

After attempting various methods of function checking, we have come to the conclusion that there is no robust method of verifying two functions are the same EXCEPT manually vetting all inputs through two functions to ensure equivalent outputs. Even this method is not perfect, and only raises confidence in sameness from the given inputs.

The solution to LazySimResult.extend() fcn checking is therefore reminding the user in code comments that a LazySimResult extending another MUST have the same fcn function.

Testing was done by creating two nearly identical functions performing the same simple task differently. As demonstrated in closed PR#209 and branch 'feature/LazySimResult_other_fcn_check', the following methods fail in properly identifying matching functions:
- Byte code via Bytecode class disassembly (Bytecode.dis(), Bytecode.info, Bytecode.codeobj)
- Byte code via obj.code.co_code
- inspect getsource()
- dis.dis

Issue #218, opened 2022-02-23
Dataset Example update


teubert:
Update example for dataset to match matlab example

Issue #217, opened 2022-02-23
Test datasets


teubert:
Add test of datasets that only runs on release candidate (i.e., PR into master)

Issue #216, opened 2022-02-23
Dataset: Other


teubert:
Add access to other datasets

Issue #215, opened 2022-02-23
Dataset: CMAPPS


teubert:
Add access to CMAPPS dataset

PR #219, opened 2022-02-23
Battery Dataset

### LINKED ISSUE #213: Consider datasets
Consider including some of the datasets (like keras does)
E.g, battery dataset 

E.g.,
From prog_models.datasets import nasa_batt
data = nasa_batt.load_data()

teubert:
Closes #213 

Added submodule for datasets. As a first dataset implemented access to battery dataset. The data is restructured to a useful format. The format chosen was the following:

* description [dictionary]
* data: List[np.Array] of format [run_id][time_index][variable] - where each run is a rank-2 tensor. This format was selected because it's more or less the format expected in ML operations for time series. 

Also, added an example demonstrating how to use it. 

NOTE: This does not include a test, that's on purpose because it takes some time

Follow-up issues created #218, #217, #216, #215, #214

### DIFF
--- a/examples/dataset.py
+++ b/examples/dataset.py
@@ -51,7 +51,18 @@ def run_example():
     plt.plot(data[4][:, 0], data[4][:, 2])
     plt.ylabel('Voltage (V)')
     plt.xlabel('Time (s)')
+    plt.title('Run 4')
+
+    # Graph all reference discharge profiles
+    indices = [i for i, x in enumerate(desc['runs']) if 'reference discharge' in x['desc'] and 'rest' not in x['desc']]
+    plt.figure()
+    for i in indices:
+        plt.plot(data[i][:, 0], data[i][:, 2], label=f"Run {i}")
+    plt.title('Reference discharge profiles')
+    plt.xlabel('Time (s)')
+    plt.ylabel('Voltage (V)')
     plt.show()
+
 
 # This allows the module to be executed directly 
 if __name__=='__main__':
--- a/examples/dataset.py
+++ b/examples/dataset.py
@@ -34,7 +34,7 @@ def run_example():
     #    1: current (amps)
     #    2: voltage
     #    3: temperature (°C)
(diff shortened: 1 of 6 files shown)


examples/dataset.py
```diff
@@ -0,0 +1,69 @@
...
+    print(f'Procedure: {desc["procedure"]}')
+
+    # Step 3: Access the dataset data
+    # Data is in format [run_id][time][variable]
+    # For the battery the variables are 
+    #    0: relativeTime (since beginning of run)
+    #    1: current (amps)
+    #    2: voltage
+    #    3: temperature (°C)
+    # so that data[a][b, 3] is the temperature at time index b (relative to the start of the run) for run 1
```
kjjarvis, line 37: Should line 37 be ... # so that data[a][b][3] is the temperature at time index b (relative to the start of the run) for run a ... The changes here are 1) the three brackets instead of two with the comma to match format [run_id][time][variable], and 2) the "run 1" in the original comment is changed to "a" to match the notation ... if I'm understanding correctly!

kjjarvis (approved):
This is a great new feature. Runs successfully and looks good

matteocorbetta (approved):
Data format is good.
Latest changes are good.

tests/test_examples.py
```diff
@@ -38,6 +40,8 @@ def run_tests():
 def main():
     # Create tests for each example
     for _, name, _ in pkgutil.iter_modules(['examples']):
+        if name in examples_skipped:
+            continue
         test_func = make_test_function(name)
         setattr(TestExamples, 'test_{0}'.format(name), test_func)
```
lawrence-hwang, line 46:
```suggestion
        if name not in examples_skipped:
            test_func = make_test_function(name)
            setattr(TestExamples, 'test_{0}'.format(name), test_func)   
```

Accidentally committed the change instead of suggesting it earlier. I ran simpler code under this control flow as well as my suggestion and saw some performance improvement in this change. Might be worth using for slightly increased efficiency?

teubert: Good catch!

PR #225, opened 2022-02-28
Matrix Model

teubert:


lawrence-hwang (approved):
Looking good!

teubert:
Decided not to go with this because of the added complexity

PR #228, opened 2022-03-03
Matrix Model Approach 2: Container

teubert:
Created a container for input, states, outputs where data is stored as a matrix, but can be accessed and manipulated as either a dictionary or a matrix. 

Also added tests

Note: This change breaks vectorized behavior, so the example was removed and issue #236 was added to resolve this.

Follow up issues (i.e., next steps):
#232 
#233 
#235 
#234 
#236

### DIFF
--- a/examples/matrix_model.py
+++ b/examples/matrix_model.py
@@ -53,8 +53,9 @@ def run_example():
 
             A = np.array([[0, 1], [0, 0]])  # State transition matrix
             B = np.array([[0], [self.parameters['g']]])  # Acceleration due to gravity
+            x.matrix += (np.matmul(A, x.matrix) + B) * dt
 
-            return self.StateContainer(x.matrix + (np.matmul(A, x.matrix) + B) * dt)
+            return x.matrix
             
         def output(self, x):
             # Note- states can still be accessed a dictionary
--- a/examples/matrix_model.py
+++ b/examples/matrix_model.py
@@ -46,6 +46,7 @@ def run_example():
                 'v': self.parameters['throwing_speed']})
 
         def next_state(self, x, u, dt):
+            # Here we will use the matrix version for each variable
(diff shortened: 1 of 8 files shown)


examples/matrix_model.py
```diff
@@ -0,0 +1,97 @@
...
+    # Use the model
+    x = thrown_object.initialize()
+    print('State at 0.1 seconds: ', thrown_object.next_state(x, {}, 0.1))
+
+    # But you can also initialize state directly, like so:
+    x = thrown_object.StateContainer({'x': 1.93, 'v': 40})
+    print('State at 0.1 seconds: ', thrown_object.next_state(x, None, 0.1))
+
+    # Now lets use it for simulation.
+    thrown_object.simulate_to_threshold(lambda t, x = None: thrown_object.InputContainer({}), print = True, threshold_keys = 'impact', dt = 0.1, save_freq = 1)
```
kjjarvis, line 93: The first two inputs to simulate_to_threshold are a bit confusing to me (I expected just future_loading and options, based on previous implementation). Do you think users would benefit from a comment on these? Or perhaps it will be in the documentation

kjjarvis (approved):
Useful implementation of extending the framework to use matrices - will be very helpful for future work

examples/matrix_model.py
teubert: Very good point. I think it would make sense not to use lambda's because they're complicated. 

```python
thrown_object.simulate_to_threshold(lambda t, x = None: thrown_object.InputContainer({})
```
is equivalent to
```python
def future_load(t, x = None):
  return thrown_object.InputContainer({})
thrown_object.simulate_to_threshold(future_load)
```

Lambda just allows you to define a simple function inline. I'll switch the syntax to something clearer

lawrence-hwang (approved):
Looks great so far!

PR #240, opened 2022-03-09
Add update function to container

teubert:
Update function will be needed by some of the existing models


lawrence-hwang (approved):
Looks good!

PR #244, opened 2022-03-11
Noise improvements

teubert:
Cleanup code by moving noise functions into separate file, using dictionary for lookup, and adding a 'none' distribution for noise. 

'none' distribution just passes the value through, so it's more efficient than setting *_noise = 0


kjjarvis (approved):
Looks good with a few (possible) changes

PR #266, opened 2022-04-21
Custom error messages for downloading + testing datasets + manual testing file

lawrence-hwang:
Add error message if the download fails (which I assume it will if it's moved) that it will print a message telling the user to upgrade to the newest version of the software because the link may have moved.

Add file for long-tests that has to be run manually. Add it to release checklist.
In long tests, include a test to make sure the datasets are downloaded properly


tests/test_manual.py
```diff
@@ -0,0 +1,76 @@
+# Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
+
```
teubert, line 2: Could you add a description to the top here in three quotes. Describe the purpose of this file, how it's not run automatically, and why

PR #288, opened 2022-05-06
Surrogate Model Tests

teubert:
Created tests for surrogate model. 

Also, added additional input validation

### DIFF
--- a/tests/test_surrogates.py
+++ b/tests/test_surrogates.py
@@ -1,6 +1,7 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
 
 import unittest
+import warnings
 
 from prog_models.models import *
 from prog_models.exceptions import ProgModelInputException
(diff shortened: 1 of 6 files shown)


tests/test_surrogates.py
```diff
@@ -0,0 +1,303 @@
...
+            self.assertEqual(surrogate_results.states[i]['x'], surrogate_results.outputs[i]['x'])
+            self.assertAlmostEqual(surrogate_results.states[i]['v'], result.states[i]['v'], delta=1)
+            self.assertAlmostEqual(surrogate_results.states[i]['falling'], result.event_states[i]['falling'], delta=0.1)
+            self.assertEqual(surrogate_results.states[i]['falling'], surrogate_results.event_states[i]['falling'])
+            self.assertAlmostEqual(surrogate_results.states[i]['impact'], result.event_states[i]['impact'], delta=0.1)
+            self.assertEqual(surrogate_results.states[i]['impact'], surrogate_results.event_states[i]['impact'])
+
+        # Events subset
+        surrogate = m.generate_surrogate([load_eqn], dt = 0.1, save_freq = 0.25,      threshold_keys = 'impact', events = ['impact'])
+        surrogate = m.generate_surrogate([load_eqn], dt = 0.1, save_freq = 0.25,      threshold_keys = 'impact', events = 'impact')
```
kjjarvis, line 210: Should both of these two lines (209 and 210) be "surrogate = "?

kjjarvis (approved):
Tests look great, very thorough, just made one comment on a potential addition and one clarifying question

lawrence-hwang (approved):
Surrogate tests look good!

tests/test_surrogates.py
teubert: One is doing it with events as a list, the other one as a string. We don't test anything except to see that no exception was raised. 

I guess there is an untested assumption that the first surrogate is identical to the second. Let me think about how to test that

tests/test_surrogates.py
```diff
@@ -0,0 +1,303 @@
...
+        with self.assertRaises(ProgModelInputException):
+            m.generate_surrogate([load_eqn], states = ['invalid'])
+        with self.assertRaises(ProgModelInputException):
+            m.generate_surrogate([load_eqn], states = ['x', 'invalid', 'v'])
+        with self.assertRaises(ProgModelInputException):
+            m.generate_surrogate([load_eqn], inputs = ['invalid'])
+        with self.assertRaises(ProgModelInputException):
+            m.generate_surrogate([load_eqn], events = ['invalid'])
+        with self.assertRaises(ProgModelInputException):
+            m.generate_surrogate([load_eqn], events = ['falling', 'impact', 'invalid'])
```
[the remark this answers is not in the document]
teubert: Wow, good catch- I'll add that

PR #289, opened 2022-05-06
Adding missing NASA copyright tags

lawrence-hwang:
Adding copyright information to files found missing the proper copyright tags.


scripts/gen_docs.sh
```diff
@@ -1,3 +1,5 @@
+# Copyright © 2022 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
```
lawrence-hwang, line 1: Is the year here correct? Added 2022 because of current year change was made.

teubert: Ah, actually #!/bin/sh needs to be the first line to work correctly.

Re: the date- excellent question. I think it actually needs to be the NASA claimed the copyright on the package- but I sent an email to the software release office to confirm

PR #296, opened 2022-05-11
Added training data noise capability for SurrogateModel

kjjarvis:
We have determined that any linear dependence between states/inputs/outputs/event states causes instability in the DMD surrogate model. We can adjust for this by including a small amount of noise in the training data

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -1147,7 +1147,9 @@ class PrognosticsModel(ABC):
             'inputs': self.inputs,
             'outputs': self.outputs,
             'events': self.events,
-            'stability_tol': 1e-05
+            'stability_tol': 1e-05,
+            'data_add_noise': True,
+            'data_noise_magnitude': 1e-02
         }
         config.update(kwargs)
 
@@ -1184,9 +1186,13 @@ class PrognosticsModel(ABC):
             for state_name in self.states:
                 states_data_temp = [states[iter_data1][state_name] for iter_data1 in range(len(states))]
                 states_data_interp[state_name] = interp1d(times,states_data_temp)(time_data_interp)
+                if config['data_add_noise'] is True:
+                    states_data_interp[state_name] = states_data_interp[state_name] + np.random.randn(len(states_data_interp[state_name]))*config['data_noise_magnitude']
             for input_name in self.inputs:
                 inputs_data_temp = [inputs[iter_data4][input_name] for iter_data4 in range(len(inputs))]
                 inputs_data_interp[input_name] = interp1d(times,inputs_data_temp)(time_data_interp)
+                if config['data_add_noise'] is True:
+                    inputs_data_interp[input_name] = inputs_data_interp[input_name] + np.random.randn(len(inputs_data_interp[input_name]))*config['data_noise_magnitude']
 
             states_data = [
                 self.StateContainer({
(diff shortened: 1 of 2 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -1147,7 +1147,9 @@ def generate_surrogate(self, load_functions, first_output = None, threshold_keys
             'inputs': self.inputs,
             'outputs': self.outputs,
             'events': self.events,
-            'stability_tol': 1e-05
+            'stability_tol': 1e-05,
+            'data_add_noise': True,
```
teubert, line 1151:
```suggestion
            'data_add_noise': True,
```

Rather than having 2 parameters, what if we just have one (magnitude) and if it's set to 0 there is no noise?

src/prog_models/prognostics_model.py
```diff
@@ -1147,7 +1147,9 @@ def generate_surrogate(self, load_functions, first_output = None, threshold_keys
...
+            'data_noise_magnitude': 1e-02
```
teubert, line 1152: I'm not sure I like the name data_noise_magnitude. What about 'training_noise'?

src/prog_models/prognostics_model.py
```diff
@@ -1184,9 +1186,13 @@ def generate_surrogate(self, load_functions, first_output = None, threshold_keys
             for state_name in self.states:
                 states_data_temp = [states[iter_data1][state_name] for iter_data1 in range(len(states))]
                 states_data_interp[state_name] = interp1d(times,states_data_temp)(time_data_interp)
+                if config['data_add_noise'] is True:
+                    states_data_interp[state_name] = states_data_interp[state_name] + np.random.randn(len(states_data_interp[state_name]))*config['data_noise_magnitude']
             for input_name in self.inputs:
                 inputs_data_temp = [inputs[iter_data4][input_name] for iter_data4 in range(len(inputs))]
                 inputs_data_interp[input_name] = interp1d(times,inputs_data_temp)(time_data_interp)
+                if config['data_add_noise'] is True:
+                    inputs_data_interp[input_name] = inputs_data_interp[input_name] + np.random.randn(len(inputs_data_interp[input_name]))*config['data_noise_magnitude']
```
teubert, line 1195:
```suggestion
                    inputs_data_interp[input_name] += np.random.randn(len(inputs_data_interp[input_name]))*config['data_noise_magnitude']
```

teubert (changes requested):
Great feature, a few suggestions

teubert (approved):

src/prog_models/prognostics_model.py
matteocorbetta: Why "data_add_noise" and not simply "add_noise"?
is there a reason for "data" to be there?

src/prog_models/prognostics_model.py
matteocorbetta: same as above.

src/prog_models/prognostics_model.py
matteocorbetta: for future implementations we may think of giving the option to add noise to the states only and not the input.
for example:
'state_add_noise': True,
'input_add_noise': False,
so that the noise is not automatically added to the input if not necessary.
However, I'm not convinced should go in right away. we need maybe a little more testing.

matteocorbetta (approved):
Very minor comments but otherwise good to go!

src/prog_models/prognostics_model.py
kjjarvis: Great point, I ended up putting these together into "training_noise"

kjjarvis: I think this is a great point. Between this suggestion and our other discussions about specifying individual state noise, etc., I think we can improve this in the future.

teubert (changes requested):
Needs test for training_noise feature

PR #298, opened 2022-05-12
Config Comparing Table

lawrence-hwang:
New Feature: Generate a table to compare the configuration of multiple models of the same type.

### DIFF
--- a/src/prog_models/utils/parameters.py
+++ b/src/prog_models/utils/parameters.py
@@ -122,3 +122,12 @@ class PrognosticsModelParameters(UserDict):
             updates = callback(self[key])
             self.update(updates)
 
+    def compare(self, *args):
+        """Generate a table to compare the configuration of multiple models of the same type.
+        
+        Args:
+            *args (PrognosticsModelParameters):
+                Additional model parameters to compare this PrognosticsModelParameters against.
+        """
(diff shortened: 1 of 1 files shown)


lawrence-hwang:
**LEAVING OFF NOTES**

This issue and PR are unfinished due to not having enough time to complete before the end of my internship with the Diagnostics and Prognostics group. In this comment, I will convey the work I have done towards a solution and where I left off in hopes that this issue will be picked up and completed at a later date.

**Completed**
* A table is represented by a list of strings, where each string is a row. The 0th row is the first line of the table, and the ith string the ith row. Tables are formulated by iterating over attributes and appending new rows to the table, which is then finally printed at the end using something like print(*table_list, sep='\n')
* Method created in PrognosticsModelParameters of parameters.py, compare, that takes any number of other parameter objects with attributes to print
* Testing structure created in test_base_models.py, unittest test_config_table()
* Elementary table printing logic implemented, currently only printing rows with attributes

**Recommended Next Steps**
* Format the table in a prettier way. Consider looking at our describe implementation and table utils over in prog_algs UncertainData
* Some attributes are stored as dictionaries. Consider printing separate table for these. Suggest accomplishing this recursively.

PR #301, opened 2022-05-12
Update datasets to use pandas dataframe

teubert:


matteocorbetta:
It looks good. Checked both battery and cmapss.
FYI, the temperature sensor of battery #2 was off or broken for the first part of the dataset (temperature values -4000), which is something that some people may notice when exploring the data.
Since it's always been like this, I don't know if it's worth it to write a Note in the nasa_battery help stating this. Up to you!
Thanks

PR #312, opened 2022-06-03
Automatic step size

teubert:
Add feature where step size can be determined automatically to capture save_pts and save_freq. This will later support using the step size in the future load as the sim step size. 

Now when setting dt you have the following options:
* Number - constant dt step size
* Function (t, x) -> dt - step size by function
* string - step size mode ('auto' or 'constant').
   * constant is equivalent to dt=1.0
   * auto will set the dt to capture any save pts and save_freq perfectly, and later the future_loading
* tuple (mode, dt)
   * ('constant', dt) - is equivalent to dt=dt
   * ('auto', dt) - sets step size automatically with an upper limit of dt. 

Set default as auto with maximum step size of 1.0.

### DIFF
--- a/examples/sim.py
+++ b/examples/sim.py
@@ -48,7 +48,7 @@ def run_example():
     options['save_freq'] = 201  # Save every 201 seconds
     options['save_pts'] = [250, 772, 1023]  # Special points we sould like to see reported
     simulated_results = batt.simulate_to_threshold(future_loading, **options)
-    # Note that event though the step size is 2, the odd points in the save frequency are met perfectly, dt is adjusted automatically to capture the save points
+    # Note that even though the step size is 2, the odd points in the save frequency are met perfectly, dt is adjusted automatically to capture the save points
 
 # This allows the module to be executed directly 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


examples/sim.py
```diff
@@ -43,6 +43,13 @@ def future_loading(t, x=None):
     }
     simulated_results = batt.simulate_to_threshold(future_loading, **options)
 
+    # Alternately, you can set a max step size and allow step size to be adjusted automatically
+    options['dt'] = ('auto', 2)  # set step size automatically, with a max of 2 seconds
+    options['save_freq'] = 201  # Save every 201 seconds
+    options['save_pts'] = [250, 772, 1023]  # Special points we sould like to see reported
+    simulated_results = batt.simulate_to_threshold(future_loading, **options)
+    # Note that event though the step size is 2, the odd points in the save frequency are met perfectly, dt is adjusted automatically to capture the save points
```
kjjarvis, line 51: "even", not "event"

kjjarvis (approved):
Great new feature for time-step flexibility

examples/sim.py
teubert:
```suggestion
    # Note that even though the step size is 2, the odd points in the save frequency are met perfectly, dt is adjusted automatically to capture the save points
```

PR #318, opened 2022-06-06
Exception when dataset has moved

teubert:
The website is currently down, and NASA had it forward to a different site. This wasn't caught by our check and the feature gave a weird error. 

This update will catch the issue and give an accurate error.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -24,7 +24,7 @@ class TestDatasets(unittest.TestCase):
         # Legit website, but it's not the repos
         nasa_cmapss.URL = "https://github.com/nasa/prog_models"
         with self.assertRaises(ConnectionError):
-            (desc, data) = nasa_cmapss.load_data(1)
+            (train, test, results) = nasa_cmapss.load_data(1)
     # Testing for successful download located in manual testing files; test_manual.py
 
 # This allows the module to be executed directly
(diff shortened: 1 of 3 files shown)


ajsharda17 (approved):

tests/test_datasets.py
```diff
@@ -7,16 +7,24 @@ class TestDatasets(unittest.TestCase):
...
     def test_nasa_cmapss_bad_url_download(self):
         from prog_models.datasets import nasa_cmapss
         BAD_URL = "https://"+"BADURLTEST"
         nasa_cmapss.URL = BAD_URL
         with self.assertRaises(ConnectionError):
             (train, test, results) = nasa_cmapss.load_data(1)
+        # Legit website, but it's not the repos
+        nasa_cmapss.URL = "https://github.com/nasa/prog_models"
+        with self.assertRaises(ConnectionError):
+            (desc, data) = nasa_cmapss.load_data(1)
```
mollocon44, line 27: Are you not expecting a tuple from cmapps like in line 23?

mollocon44 (changes requested):
Noticed a difference in the return values expected between line 23 and the rest of the test cases - sorry if this is okay and my python is just rusty

tests/test_datasets.py
teubert:
```suggestion
            (train, test, results) = nasa_cmapss.load_data(1)
```
good catch!

mollocon44 (approved):

PR #322, opened 2022-06-13
Feature/surrogate noise

kjjarvis:
New feature to allow noise to be added to surrogate modeling training data, to eliminate any quasi-linear dependence among states that can result in instabilities

### DIFF
--- a/src/prog_models/surrogates/dmd.py
+++ b/src/prog_models/surrogates/dmd.py
@@ -31,6 +31,8 @@ class SurrogateDMDModel(LinearModel):
             Note: To trim data to a set time, use the 'horizon' parameter\n   
         stability_tol: int, optional
             Value that determines the tolerance for DMD matrix stability\n
+        training_noise: int, optional
+            Noise added to the training data sampled from a standard normal distribution with magnitude training_noise \n
 
     See Also
     ---------
@@ -64,7 +66,8 @@ class SurrogateDMDModel(LinearModel):
         # Configure
         config = { # Defaults
             'trim_data_to': 1,
-            'stability_tol': 1e-05
+            'stability_tol': 1e-05,
+            'training_noise': 1e-05
         }
         config.update(kwargs)
 
@@ -72,6 +75,8 @@ class SurrogateDMDModel(LinearModel):
             raise ProgModelInputException("Invalid 'trim_data_to' input value, must be between 0 and 1.")
         if not isinstance(config['stability_tol'], Number) or  config['stability_tol'] < 0:
             raise ProgModelInputException(f"Invalid 'stability_tol' input value {config['stability_tol']}, must be a positive number.")
+        if not isinstance(config['training_noise'], Number) or config['training_noise'] < 0:
+            raise ProgModelInputException(f"Invalid 'training_noise' input value {config['training_noise']}, must be a positive number.")
 
         states_dmd = m.states.copy()
         inputs_dmd = m.inputs.copy()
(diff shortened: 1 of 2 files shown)


src/prog_models/surrogates/dmd.py
```diff
@@ -31,6 +31,8 @@ class SurrogateDMDModel(LinearModel):
             Note: To trim data to a set time, use the 'horizon' parameter\n   
         stability_tol: int, optional
             Value that determines the tolerance for DMD matrix stability\n
+        training_noise: int, optional
+            Noise added to the training data sampled from a standard normal distribution with magnitude training_noise \n
```
teubert, line 35:
```suggestion
            Noise added to the training data sampled from a standard normal distribution with standard deviation of training_noise \n
```

tests/test_surrogates.py
```diff
@@ -258,6 +263,145 @@ def load_eqn(t = None, x = None):
...
+        }
+
+        surrogate_results = surrogate.simulate_to_threshold(load_eqn, **options)
+        surrogate_noise_results = surrogate_noise.simulate_to_threshold(load_eqn, **options)
+
+        position = [surrogate_results.states[iter]['x'] for iter in range(len(surrogate_results.times))]
+        position_noise = [surrogate_noise_results.states[iter]['x'] for iter in range(len(surrogate_noise_results.times))]
+
+        MSE = sum([(position[iter] - position_noise[iter])**2 for iter in range(min(len(position),len(position_noise)))])/min(len(position),len(position_noise))
+        self.assertLess(MSE, 10) #
```
teubert, line 293:
```suggestion
        self.assertLess(MSE, 10)
        self.assertNotEqual(MSE, 0)  # Shouldn't be exact
```

teubert: Same with other examples below

teubert (changes requested):
Your tests are great! 

One comment though. Your tests make sure that the noise don't cause the results to go off by too much (i.e., that they're about equal), but we should also make sure the results are not exactly equal (for example, if there's a bug that causes noise to not work at all).

tests/test_surrogates.py
kjjarvis: very good idea! Adding some more tests now

teubert (approved):

ArjunSharda (approved):

src/prog_models/surrogates/dmd.py
```diff
@@ -94,31 +99,43 @@ def __init__(self, m, load_functions, **kwargs):
...
+            for output_name in m.outputs:
+                outputs_data_temp = [outputs[iter_data3][output_name] for iter_data3 in range(len(outputs))]
+                outputs_data_interp[output_name] = interp1d(times,outputs_data_temp)(time_data_interp)
+                if config['training_noise'] != 0:
+                    outputs_data_interp[output_name] += np.random.randn(len(outputs_data_interp[output_name]))*config['training_noise']
+            for es_name in m.events:
+                es_data_temp = [event_states[iter_data2][es_name] for iter_data2 in range(len(event_states))]
+                es_data_interp[es_name] = interp1d(times,es_data_temp)(time_data_interp)
+                if config['training_noise'] != 0:
+                    es_data_interp[es_name] += np.random.randn(len(es_data_interp[es_name]))*config['training_noise']
```
matteocorbetta, line 119: avoid if? Just use:
states_data_interp[state_name] += np.random.randn(len(states_data_interp[state_name]))*config['training_noise']

matteocorbetta (approved):

src/prog_models/surrogates/dmd.py
teubert: That will remove a line of code, which is nice. But the way Katy implemented it now will be much more efficient when time is 0. 

I timed it to illustrate. 
Case 1: training_noise is 0
   with if statement: 0.029 micro seconds per event
   without if statement: 3.8 micro seconds per event
Case 2: training_noise is not 0
   with if statement: 3.96 micro seconds per event
    without if statement: 3.8 micro seconds per event
    
Since it's not in the inner loop of simulation - I would accept it either way. It will save 10s of micro seconds in creating the surrogate if there's no noise, while only adding ~1s of microseconds if there is noise. I don't think it hurts to have this micro optimization.

teubert: Here's my code if anyone wants to recreate it

`
>>> import time
>>> f1 = """\
... a += np.random.randn(len(a))*b
... """
>>> f2 = """\
... if b != 0:
...    a += np.random.randn(len(a))*b
... """
>>> timeit.timeit(f1, setup="import numpy as np; b=0; a = np.array([[1.0, 2.2],[2.1, 3.1]])")
>>> timeit.timeit(f2, setup="import numpy as np; b=0; a = np.array([[1.0, 2.2],[2.1, 3.1]])")
>>> timeit.timeit(f1, setup="import numpy as np; b=1; a = np.array([[1.0, 2.2],[2.1, 3.1]])")
>>> timeit.timeit(f2, setup="import numpy as np; b=1; a = np.array([[1.0, 2.2],[2.1, 3.1]])")
`

kjjarvis: Really interesting results! Thanks for running that. I agree that both options have pros and cons based on speed and likely frequency. For simplicity sake, I think I'll leave it as-is for now, and we'll gain the small micro-optimization. Thanks to both of you for your perspective on this.

matteocorbetta (commented):
ok thanks for checking the 'if' statement stuff!

PR #324, opened 2022-07-06
Various Optimizations

teubert:
Various optimizations that were uncovered through benchmarking with @kjjarvis, including:
* Removing duplicate deepcopying in constructing simresult (deep copy state to put in array, then again in simresult constructor). Fixed by appending to empty simresult instead of list.
* Add explicit threshold_met for linear model (removes call to event_state and dict reformulation


teubert:
I ran profiling on my computer and this change showed a 33% improvement!

PR #363, opened 2022-08-05
Refactored lstm 

teubert:
Refactored LSTM into a subpackage data_models, and added abstract base class DataModel. 

Also, set default noise to 0

This is in anticipation of future data-driven models, and adding general tools. The change formalizes the interface. Future changes will also bring in DMD Model as a subclass

### DIFF
--- a/examples/custom_model.py
+++ b/examples/custom_model.py
@@ -1,0 +1,117 @@
+# Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
+# This ensures that the directory containing examples is in the python search directories
+
+"""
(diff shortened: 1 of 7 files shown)


examples/custom_model.py
```diff
@@ -0,0 +1,117 @@
+# Copyright © 2021 United States Government as represented by the Administrator of the National Aeronautics and Space Administration.  All Rights Reserved.
```
AliPohya, line 1: On one of these slow days, we could search and replace the 2021 copyright statement and update it to 2022.

teubert: Actually I was told by the lawyers to keep the 2021 copyright year because that's when we received permission to first release. 

It surprised me too

AliPohya (approved):

Issue #361, opened 2022-08-05
Move DMD to data_models class


teubert:
Follows #345

PR #365, opened 2022-08-08
Add DataModel.from_model

teubert:
Added from_model class method to DataModel. This will now allow users to create models from others (i.e., surrogates) for any data model class. It essentially generates data from the original model, and then calls the from_data method.

### Made a few changes to the from_data method to support this:
* Split data into inputs, outputs, states, and event_states. This allows model classes to all accept the data they need (e.g., DMD needs all these, but LSTM only needs the first 2). Any additional information received will be ignored
* Renamed the previous inputs, outputs, etc. to input_keys, output_keys, etc. This is to not conflict with the inputs/outputs from the above bullet
* Added a few additional input validation checks into LSTMStateTransitionModel.from_data
* Moved PrognosticsModel.events to be object member instead of class member. This is to prevent a weird behavior where we add to events/inputs/outputs in building the model by appending from the original, and that action editing the default array.  
* Update PrognosticsModel to allow empty input. Not related, just occurred to me as I was doing the above change
* Updated DictLikeMatrixWrapper to allow subset of keys (others are replaced with None). This made building future_loading models for surrogates using the full model easier. 

### A few questions for reviewers (beyond the usual):
* Is the naming inputs, input_keys clear? should we instead do input_data and input_keys?
* Does the interface make sense?

### Follow-on issues for future PRs:
 * #361

### DIFF
--- a/tests/test_lstm.py
+++ b/tests/test_lstm.py
@@ -144,9 +144,9 @@ class TestLSTM(unittest.TestCase):
     def test_improper_input(self):
         # Input, output format
         with self.assertRaises(Exception, msg='No outputs'):
-            LSTMStateTransitionModel.from_data(inputs=[])
+            LSTMStateTransitionModel.from_data(inputs=[1])
         with self.assertRaises(Exception, msg='No inputs'):
-            LSTMStateTransitionModel.from_data(outputs=[])
+            LSTMStateTransitionModel.from_data(outputs=[1])
         with self.assertRaises(ValueError, msg='Empty inputs/outputs'):
             LSTMStateTransitionModel.from_data(inputs=[], outputs=[])
         with self.assertRaises(ValueError, msg="Uneven Inputs/outputs"):
--- a/tests/test_lstm.py
+++ b/tests/test_lstm.py
@@ -21,10 +21,11 @@ class TestLSTM(unittest.TestCase):
 
         # Step 2: Generate model
         m2 = LSTMStateTransitionModel.from_data(
-            (data.inputs, data.outputs),  
+            inputs = [data.inputs],
+            outputs = [data.outputs],  
             window=5, 
-            epochs=50,
-            outputs = ['x'])  
+            epochs=25,
+            output_keys = ['x'])  
         self.assertIsInstance(m2, LSTMStateTransitionModel)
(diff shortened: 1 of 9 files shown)


tests/test_lstm.py
```diff
@@ -54,49 +55,149 @@ def future_loading2(t, x = None):
...
-            LSTMStateTransitionModel.from_data(12)
-        with self.assertRaises(ValueError, msg="Empty Tuple"):
-            LSTMStateTransitionModel.from_data(())
-        with self.assertRaises(ValueError, msg="Small Tuple"):
-            LSTMStateTransitionModel.from_data((1))
-        with self.assertRaises(ValueError, msg="Too large Tuple"):
-            LSTMStateTransitionModel.from_data((1, 2, 3))
+        # Input, output format
+        with self.assertRaises(Exception, msg='No outputs'):
+            LSTMStateTransitionModel.from_data(inputs=[])
```
kjjarvis, line 147: Is this line flipped with line 149? I'm not sure I completely understand the syntax of the testing here, but I think lines 146-147 are testing what happens when "inputs = []" and this raises the exception "No outputs" ... should these be the same (either inputs or outputs, with the other tested in 148-149)

kjjarvis (approved):
Nice new interface for data-driven models. I like the use of input_data and input_keys, I think it's clear throughout.

tests/test_lstm.py
teubert: Actually this is right. You're right that line 147 corresponds with the check on 146. In line 147 inputs are provided (an empty array) but no outputs at all. I can see how this would be confusing since I'm essentially providing no inputs since the array is empty. I think it will cause less confusion if I put a number in there. 

```suggestion
            LSTMStateTransitionModel.from_data(inputs=[1])
```

tests/test_lstm.py
```diff
@@ -54,49 +55,149 @@ def future_loading2(t, x = None):
...
+        with self.assertRaises(Exception, msg='No inputs'):
+            LSTMStateTransitionModel.from_data(outputs=[])
```
teubert, line 149:
```suggestion
            LSTMStateTransitionModel.from_data(outputs=[1])
```

PR #368, opened 2022-08-08
SimResult to numpy array

teubert:
Implement method to convert from simresult to numpy array. This will be necessary for some of the data-driven approaches.


hlembo (approved):
Looks good to me!

PR #376, opened 2022-08-12
Make DataModels Pickle-able

teubert:
This PR will ensure that DataModels (e.g., LSTM and DMD) can be pickeled and restored. To do so we define the __setstate__ and __getstate__ methods which are called to get the configuration and configure the constructed object. We also define the __getnewargs__ method, which defines the arguments for the __new__ method, when defined. 

To do this, I also had to make DMD matrix an attribute of the DMD class

Also, in DMD.__init__, reordered params building and input checking to make the exception if you give incorrect input a TINY bit faster

### DIFF
--- a/tests/test_data_model.py
+++ b/tests/test_data_model.py
@@ -9,32 +9,7 @@ from prog_models.models import ThrownObject
 import sys
 from io import StringIO
 
-class TestDataModel(unittest.TestCase):
-    def _model_tests(self, m, m2, DataModelType, max_error, TIMESTEP, data):
-        self.assertIsInstance(m2, DataModelType)
-        self.assertIsInstance(m2, DataModel)
-        self.assertListEqual(m2.outputs, list(m.outputs))
-        
-        # Step 3: Use model to simulate_to time of threshold
-        t_counter = 0
-        x_counter = m.initialize()
-        def future_loading2(t, x = None):
-            # Future Loading is a bit complicated here 
-            # Loading for the resulting model includes the data inputs, 
-            # and the output from the last timestep
-            nonlocal t_counter, x_counter
-            z = m.output(x_counter)
-            z = m2.InputContainer(z.matrix)
-            x_counter = m.next_state(x_counter, future_loading(t), t - t_counter)
-            t_counter = t
-            return z
-        
-        results2 = m2.simulate_to(data.times[-1], future_loading2, dt=TIMESTEP, save_freq=TIMESTEP)
-
-        # Have to do it this way because the other way (i.e., using the LSTM model- the states are not a subset)
-        # Compare RMSE of the results to the original data
-        error = m.calc_error(results2.times, results2.inputs, results2.outputs)
-        self.assertLess(error, max_error)
+class TestDataModel(unittest.TestCase):        
 
     def _test_simple_case(self, 
         DataModelType, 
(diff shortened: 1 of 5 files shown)


tests/test_data_model.py
```diff
@@ -71,13 +71,30 @@ def test_lstm_simple(self):
         # Create from model
         LSTMStateTransitionModel(m.model, output_keys = ['x'])
 
+        # Test pickling model m
+        import pickle
```
darrahts, line 75: Suggest to move to the top of the file.

tests/test_data_model.py
```diff
@@ -71,13 +71,30 @@ def test_lstm_simple(self):
...
     def test_dmd_simple(self):
-        self._test_simple_case(DMDModel, max_error=4)
+        self._test_simple_case(DMDModel, max_error=6)
 
         # Without velocity, DMD doesn't perform well
-        self._test_simple_case(DMDModel, WITH_STATES = False, max_error=100)
+        m = self._test_simple_case(DMDModel, WITH_STATES = False, max_error=100)
+
+        # Test pickling model m
+        import pickle
```
darrahts, line 91: Suggest to move to the top of the file.

tests/test_data_model.py
darrahts: This could be a Windows issue, but I also know there are caveats with pickling keras models and I actually have never pickled a keras model - always used the built in functions. 
 
I get the following error for test_lstm_simple:

```
======================================================================
ERROR: test_lstm_simple (__main__.TestDataModel)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "<ipython-input-1-e2dee28b68ec>", line 79, in test_lstm_simple
    m2 = pickle.loads(pickled_m)
  File "C:\Users\darrahts\anaconda3\envs\tf2x\lib\site-packages\keras\saving\pickle_utils.py", line 48, in deserialize_model_from_bytecode
    model = save_module.load_model(temp_dir)
  File "C:\Users\darrahts\anaconda3\envs\tf2x\lib\site-packages\keras\utils\traceback_utils.py", line 67, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "C:\Users\darrahts\anaconda3\envs\tf2x\lib\site-packages\tensorflow\python\saved_model\load.py", line 978, in load_internal
    str(err) + "\n You may be trying to load on a different device "
FileNotFoundError: Unsuccessful TensorSliceReader constructor: Failed to find any matching files for ram://5766af8a-251b-4258-a347-f22418e2ada6/variables/variables
 You may be trying to load on a different device from the computational device. Consider setting the `experimental_io_device` option in `tf.saved_model.LoadOptions` to the io_device such as '/job:localhost'.

----------------------------------------------------------------------
Ran 4 tests in 39.067s

FAILED (errors=1)
```

references:

https://github.com/tensorflow/tensorflow/issues/34697
https://stackoverflow.com/questions/73294899/unable-to-load-tensorflow-model-with-pickle

tests/test_data_model.py
```diff
@@ -71,13 +71,30 @@ def test_lstm_simple(self):
         # Create from model
         LSTMStateTransitionModel(m.model, output_keys = ['x'])
 
+        # Test pickling model m
+        import pickle
+        pickled_m = pickle.dumps(m)
+        m2 = pickle.loads(pickled_m)
```
darrahts, line 77: Just to verify the above, I changed lines 76 and 77 to 
```
pickled_m = pickle.dump(m, open('model.pkl', 'wb')) # dumps --> dump
m2 = pickle.load(open('model.pkl', 'rb')) # loads --> load
```
and received the same error.

tests/test_data_model.py
```diff
@@ -71,13 +71,30 @@ def test_lstm_simple(self):
...
+        m = self._test_simple_case(DMDModel, WITH_STATES = False, max_error=100)
+
+        # Test pickling model m
+        import pickle
+        pickled_m = pickle.dumps(m)
+        m2 = pickle.loads(pickled_m)
+        self.assertIsInstance(m2, DMDModel)
+        self.assertIsInstance(m2, DataModel)
+        self.assertListEqual(m2.outputs, ['x'])
+
```
darrahts, line 97: Test passes, but the following warning is generated:

```
UserWarning: State value 'x' is duplicated in inputs, outputs, or events; duplicate has been removed.
```

Nothing stands out immediately that causes this (likely can be ignored?)

darrahts (changes requested):

tests/test_data_model.py
teubert: Very interested- it seems to work fine on Mac and Linux. 

Since pickling doesn't seem to work well with keras models, I added a warning which should now come up when anyone tries to pickle the LSTMStateTransitionModel

tests/test_data_model.py
teubert: That warning's expected. DMD is trying to be smart and eliminate duplicate keys.

darrahts:
The failed check can be ignored, all looks good!

PR #384, opened 2022-08-23
Added method to plot history in LSTMStateTransitionModel

teubert:
This PR Makes a few changes to the LSTMStateTransitionModel:
1. Exposes the model fitting history to the user (as m.history)
2. Adds method to plot the error vs epoch in training
3. Updates examples to plot the error (from 2 above)


matteocorbetta (approved):
Looks good to me as a first step.
A good thing would be to add training and validation metrics in the same plot.
For example:
training loss and validation loss 
training RMSE and validation RMSE
because that would help the user understand if the training is working well (e.g., overfitting).
maybe next iteration?

teubert:
> Looks good to me as a first step.
> A good thing would be to add training and validation metrics in the same plot.
> For example:
> training loss and validation loss
> training RMSE and validation RMSE
> because that would help the user understand if the training is working well (e.g., overfitting).
> maybe next iteration?

Good idea! I opened an issue for this. See #392

PR #389, opened 2022-08-25
Make initialize optional

teubert:
This PR makes the initialize function optional- defaulting to returning parameters['x0'] - which is how most models would do it. 

This simplifies things for users. Some users were confused by the intialize function


hlembo (approved):
Looks Good.

Issue #390, opened 2022-08-26
insufficient check for file type


snoyes:
https://github.com/nasa/prog_models/blob/103b6c388ac00edbcf6631c98872a53898947dfe/scripts/test_copyright.py#L17

Checking that the last two characters are 'py' is not sufficient for finding Python files.

Other extensions may end in 'py'. For example, the concept-tagging-api repository includes some '.npy' files, which are NumPy data files. They will not have the license line.

Python files might not end in 'py'. For example, the EMTG repository includes a '.pyw' file, which is used in Windows to prevent opening a new terminal window.

If reliance on extensions is sufficient, I recommend something like `and path.split(".")[-1] in ("py", "pyw", "py3", "ipynb")`

teubert:
Thank you @snoyes- good catch. Currently we don't have any .pyw or .py3 files, but I could see this becoming an issue in the future. 

If you have time- could you provide a review for the PR fixing this: #391?

PR #391, opened 2022-08-26
Fix test copyright

teubert:
Fix issue where our copyright testing script missed some python files with alternate file extensions (e.g., .pyw or .py3) and checked files that did not require copyright statement (e.g., .npy). 

Issue first identified by user @snoyes in #390


snoyes (changes requested):
Keep the os.path.isfile() test. It's unlikely that anyone is going to create a FIFO or named pipe with a '.py' extension, but if they did, this code would consume it.

teubert:
> Keep the os.path.isfile() test. It's unlikely that anyone is going to create a FIFO or named pipe with a '.py' extension, but if they did, this code would consume it.

Good thought- fixed

snoyes (approved):
Looks good.

PR #393, opened 2022-08-30
Split inputs, outputs from preprocessing

teubert:
Try to make the preprocessing function consistent with the other functions (e.g., from_data) by splitting out inputs, outputs. 

This also eliminates the need to create this new data list, duplicating the structures.


src/prog_models/data_models/lstm_model.py
```diff
@@ -126,23 +126,22 @@ def summary(self, file= sys.stdout, expand_nested=False, show_trainable=False):
...
         Returns:
             Tuple[ndarray, ndarray]: pre-processed data (input, output). Where input is of size (num_sequences, window, num_inputs) and output is of size (num_sequences, num_outputs)
         """
-        # Data is a List[Tuple] or Tuple (where Tuple is equivilant to [Tuple]))
-        # Tuple is (input, output)
 
         u_all = []
         z_all = []
-        for (u, z) in data:
+        for (u, z) in zip(inputs, outputs):
```
mollocon44, line 144: Do we need to make sure inputs and outputs are the same size, or will zip do that for us? All sizes moving forward are defined by the size of inputs only

mollocon44 (commented):
Only comment is on whether it's necessary to check that inputs & outputs are the same length in lstm_model.py (comment added line 144)

PR #395, opened 2022-09-08
Update LSTM to split model

teubert:
Split the LSTM model into 2 parts (state_model and output_model), in anticipation of future iterations that will also learn relationship with event_state, thresholds_met, etc. 

Designed in such a way as to still allow passing in one model, if desired.


darrahts (approved):
This now allows for multiple different outputs to share a common base model. Looks good!

PR #396, opened 2022-09-09
Early stopping

teubert:
Implement Early Stopping for LSTM Model. 

This feature, described here: https://keras.io/api/callbacks/early_stopping/, will stop the training of the epoch once it reaches certain conditions. Conditions can be configured using the early_stop.cfg parameter. 

Feature is tested by running two runs. One with early stopping, and one without. Then I parse the output to see how many epochs were run. For early stopping it should be less than the maximum.

Also, added note to example, explaining this

### DIFF
--- a/examples/lstm_model.py
+++ b/examples/lstm_model.py
@@ -43,7 +43,7 @@ def run_example():
         inputs = [data.inputs],
         outputs = [data.outputs],  
         window=4, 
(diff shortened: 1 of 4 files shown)


GGorospe (approved):
setup for the test_early_stopping function, looks like an implementation of the EarlyStopping class.

PR #397, opened 2022-09-14
Extending LSTM to include event_states, thresh_met

teubert:
Extending the divided LSTM model introduced in #395 to include estimation of threshold_met and event_state. Each of these are optional, only working if the relevant data is provided. 

Event_state is just like outputs, but thresholdmet is calculated using softmax (since it's a classification problem). 

Example for battery was extended to include this. 

[Presentation1.pdf](https://github.com/nasa/prog_models/files/9568311/Presentation1.pdf)

### DIFF
--- a/examples/custom_model.py
+++ b/examples/custom_model.py
@@ -51,7 +51,7 @@ def run_example():
 
     # Step 3: Build custom model
     print('Building custom model...')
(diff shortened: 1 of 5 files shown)


matteocorbetta (approved):
As discussed in the office. This implementation looks good so far. Need to be tested and maybe validated in other scenarios (like water filter), but looks good.

PR #399, opened 2022-09-16
Feature/tojson fromjson

kjjarvis:
Added capability to serialize (using to_json) and de-serialize (using from_json) a DMD surrogate model.


examples/serialization.py
```diff
@@ -0,0 +1,113 @@
...
+    results_serial_2 = surrogate_serial_2.simulate_to_threshold(future_loading, **options_sim)
+
+    # Plot results for comparison
+    voltage_orig = [results_orig.outputs[iter]['v'] for iter in range(len(results_orig.times))]
+    voltage_serial_1 = [results_serial_1.outputs[iter]['v'] for iter in range(len(results_serial_1.times))]
+    voltage_serial_2 = [results_serial_2.outputs[iter]['v'] for iter in range(len(results_serial_2.times))]
+
+    plt.plot(results_orig.times,voltage_orig,'-b',label='Original surrogate') 
+    plt.plot(results_serial_1.times,voltage_serial_1,'--r',label='First serialized surrogate') 
+    plt.plot(results_serial_2.times,voltage_serial_2,'-.g',label='Second serialized surrogate')
```
teubert, line 105: results_serial_2.outputs.plot(keys=['v']) instead of all this.

src/prog_models/data_models/dmd.py
```diff
@@ -388,7 +390,7 @@ def from_model(cls, m, load_functions, **kwargs):
         process_noise_temp = {key: 0 for key in m.events}
         config = {
             'add_dt': False,
-            'x0': True,  # Set it to anything not None since we define our own function
+            # 'x0': True,  # Set it to anything not None since we define our own function
```
teubert, line 393: You need to update line 195 to say:

```python
if 'x0' in self.parameters and not isinstance(self.parameters['x0'], self.StateContainer):
```

teubert (changes requested):
Great implementation and very useful feature!

I added a few comments. I'm also not sure about the from_json existing at the DMDModel level and to only at the parameters option. I think they should be at the same level for clarity. It seems that we have 2 options:

### Option 1: to/from json at model level

If model can be initialized from only the parameters, then implement these. Then you can do:
```python
json = dmd_model.to_json()
m = DMDModel.from_json(json)
```

Pros: Straightforward
Cons: Will only work for some models, requires extra work when implementign

### Option 2: parameter level

Then the model has to be implemented before importing from json. E.g., 
```python
json = dmd_model.parameters.to_json()
m = DMDModel()
m.parameters.from_json(json)
```

Pros: Will work for all models. 
Cons: A little more complicated, having to use the parameters. Have to create the model first.

examples/serialization.py
```diff
@@ -0,0 +1,113 @@
...
+    voltage_serial_2 = [results_serial_2.outputs[iter]['v'] for iter in range(len(results_serial_2.times))]
+
+    plt.plot(results_orig.times,voltage_orig,'-b',label='Original surrogate') 
+    plt.plot(results_serial_1.times,voltage_serial_1,'--r',label='First serialized surrogate') 
+    plt.plot(results_serial_2.times,voltage_serial_2,'-.g',label='Second serialized surrogate') 
+    plt.legend()
+    plt.xlabel('Time (sec)')
+    plt.ylabel('Voltage (volts)')
+    plt.show()
+
```
teubert, line 110: If we're demonstrating serialization, can we also include a note at the end saying how pickling can also be used?

examples/serialization.py
kjjarvis: Good suggestion. We decided to leave this as is for now, until we add a feature to easily put plots on the same axis using Chris' notation above

src/prog_models/utils/parameters.py
```diff
@@ -135,3 +139,29 @@ def register_derived_callback(self, key : str, callback : Callable) -> None:
...
+                save_temp['data'] = pkl_temp.decode()
+                save_temp['original_type'] = 'pickled'
+                return save_temp
+
+    def to_json(self):
+        """
+        Serialize parameters to save as JSON objects 
+        """
+
+        return json.dumps(self.data, cls=self.CustomEncoder)
```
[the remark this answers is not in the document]
kjjarvis: Great suggestion to make sure to_json and from_json are at the same level. We decided to put both at the model level. See changes in updated PR

kjjarvis:
Thanks for the great suggestions from @teubert. I've adjusted the JSON serialization to be at the model level, so now both to_json and from_json are within prognostics_model.

examples/serialization.py
```diff
@@ -0,0 +1,122 @@
+# Copyright © 2022 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration.  All Rights Reserved.
+
+"""
+Example of serializing and de-serializing a surrogate model using JSON and pickling methods
```
teubert, line 5: Question: Does this being a surrogate model add anything to the example. I wondering if we should just do the ElectroChem instead. What are your thoughts

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+                elif o['original_type'] == 'DictLikeMatrixWrapper':
+                    del o['original_type']
+                    return DictLikeMatrixWrapper(list(o.keys()),o)
+                elif o['original_type'] == 'pickled':
+                    import pickle
+                    from base64 import b64decode
+                    pkl_temp1 = o['data'].encode()
+                    pkl_temp2 = b64decode(pkl_temp1)
+                    return pickle.loads(pkl_temp2)
+            return o
```
teubert, line 1368: Let's add an else statement here to the original type check that will raise an exception. Something like f"Type {o['original_type']} not supported by PrognosticsModel json decoder

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+        parameters_dict = {}
+        for key in self.parameters.keys():
+            parameters_dict[key] = self.parameters[key]
+
+        return json.dumps(parameters_dict, cls=self.CustomEncoder)
+    
+    @classmethod
+    def from_json(cls,data):
+        """
+        Create a prognostics model model from a previously generated model that was serialized as a JSON object
```
teubert, line 1338:
```suggestion
        Create a new prognostics model from a previously generated model that was serialized as a JSON object
```

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+            data: 
+                JSON serialized parameters necessary to build a model 
+                See to_json method 
+
+        Returns:
+            PrognosticsModel: Model generated from serialized parameters 
+
+        Note
+        ----
+        This serialization only works for models that include all parameters necessary to generate the model in the parameters class.
```
teubert, line 1350:
```suggestion
        This serialization only works for models that include all parameters necessary to generate the model in the model.parameters. 
```

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+                save_temp['original_type'] = 'pickled'
+                return save_temp
+    
+    def to_json(self):
+        """
+        Serialize parameters to save as JSON objects 
+
+        Note
+        ----
+        This method only serializes the values in the parameters class of the prognostics model
```
teubert, line 1327:
```suggestion
        This method only serializes the values in the of the prognostics model parameters (model.parameters)
```

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+                from base64 import b64encode
+                pkl_temp = b64encode(pickle.dumps(o))
+                save_temp = {}
+                save_temp['data'] = pkl_temp.decode()
+                save_temp['original_type'] = 'pickled'
+                return save_temp
+    
+    def to_json(self):
+        """
+        Serialize parameters to save as JSON objects
```
teubert, line 1323:
```suggestion
        Serialize parameters as JSON objects 
```

src/prog_models/prognostics_model.py
```diff
@@ -1296,3 +1298,75 @@ def generate_surrogate(self, load_functions, method = 'dmd', **kwargs):
...
+
+    class CustomEncoder(json.JSONEncoder):
+        """
+        Custom encoder to serialize parameters 
+        """
+        def default(self, o):
+            if isinstance(o, np.ndarray):
+                return {'original_type': 'ndarray', 'data': o.tolist()}
+            elif isinstance(o, DictLikeMatrixWrapper):
+                dict_temp = {o.keys()[iter]: o[o.keys()[iter]] for iter in range(len(o.keys()))}
```
teubert, line 1310:
```suggestion
                dict_temp = {k: v for k, v in o.items()}
```

prog_model_template.py
```diff
@@ -95,6 +95,10 @@ class ProgModelTemplate(PrognosticsModel):
     def __init__(self, **kwargs):
         """
         Constructor for model
+
+        Note 
+        ----
+        To use the JSON serialization capabilities in to_json and from_json, the keyword arguments must include everything necessary for init
```
teubert, line 101:
```suggestion
        To use the JSON serialization capabilities in to_json and from_json, model.parameters must include everything necessary for initialize, including any keyword arguments.
```

examples/serialization.py
```diff
@@ -0,0 +1,122 @@
...
+    with open('surrogate_model_save_json.txt') as infile: 
+        load_surrogate_json = infile.read()
+
+    surrogate_serial_2 = DMDModel.from_json(load_surrogate_json)
+
+    # METHOD 2: Serialize by pickling
+    pickle.dump(surrogate_orig, open('surrogate_model_save_pkl.pkl','wb'))
+    surrogate_pkl = pickle.load(open('surrogate_model_save_pkl.pkl','rb'))
+
+    ## Step 5: Simulate to threshold and compare results
```
teubert, line 81:
```suggestion
    # Note: Model serialization has a lot of purposes, like saving a specific model to a file to be loaded later or sending a model to another machine over a network connection. 
    
    ## Step 5: Simulate to threshold and compare results
```

teubert (changes requested):
It's looking great. I like this setup. 

I have a few small suggestions, most are documentation changes.

examples/serialization.py
kjjarvis: I agree, much simpler with a standard battery model. Made this change.

examples/serialization.py
kjjarvis: I like this comment a lot. Is it okay to move it upwards a bit in the example, to where we first use serialization? See change in updated PR

teubert (approved):

PR #409, opened 2022-10-06
A few LSTM Bugfixes

teubert:
The LSTM model was performing poorly, resulting in poor test performance, like that shown below (right is LSTM, left is original)
MicrosoftTeams-image (2)

Investigation of this fault identified a few issues, some with the tests, others with the actual functionality. 
1. LSTM preprocessing wasn't using all the data. It was missing one datapoint for each run
2. LSTM preprocessing wasn't calculating the output window correctly (z_ii) for appending to u.
3. Tests - Training just had one run, and part of it was used for validation, so the training data was incomplete. To fix this I duplicated the data. Not good practice, but it was ok for the test. 

I also increased the default number of epochs. The default was unreasonably low.


kjjarvis (approved):
Good find on bugfixes for LSTM model

PR #411, opened 2022-10-11
Reconfigure numpy bool in serialization encoder

kjjarvis:
When serializing a model, if the custom encoder received a numpy.bool_ value, it pickled the result. We've now added a reconfiguration of these numpy.bool_ values to be standard bool values, which are supported by JSON so no pickling is required.

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -1307,7 +1307,13 @@ class PrognosticsModel(ABC):
             elif isinstance(o, DictLikeMatrixWrapper):
                 dict_temp = {k: v for k, v in o.items()}
                 dict_temp['_original_type'] = 'DictLikeMatrixWrapper'
-                return dict_temp 
+                return dict_temp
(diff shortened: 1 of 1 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -1344,7 +1344,13 @@ def default(self, o):
             elif isinstance(o, DictLikeMatrixWrapper):
                 dict_temp = {k: v for k, v in o.items()}
                 dict_temp['_original_type'] = 'DictLikeMatrixWrapper'
-                return dict_temp 
+                return dict_temp
+            elif isinstance(o,np.bool_):
+                if o == False:
```
teubert, line 1349: It might be cleaner to use casting. For example:

return bool(o) 
instead of line 1349-1353

kjjarvis: Great suggestion, made the edit

teubert (approved):

PR #414, opened 2022-10-17
Make Event State equation optional

teubert:
This PR updates behavior if event state or threshold_met is not provided:

* If event_state is not overridden, but threshold_met is: Will return event_state of 0 if threshold_met is True, else 1
* if threshold_met is overridden, but event_state is not: Will return threshold_met of True if event_state is 0, else False
* if neither are overridden: Will return empty dict (i.e., no events).


kjjarvis (approved):
Looks good, small edit suggested for clarity but not necessary for correctness

PR #437, opened 2022-11-07
Switch circuit model to use kelvin

teubert:
There are two base battery models: ElectroChemical and Circuit. They are intended to be interchangeable with each other, each have advantages over the other (ElectroChem is more accurate and complete, circuit is faster) and therefore used in different cases.

However, in practice they are not because the Circuit expects temperature in °C, whereas the electrochem expects Kelvin. To fix this I've changed the Circuit model to also use kelvin. 

This issue came up when testing the new EnsembleModel feature.


chetankul (approved):
Both the models are now using K.

PR #438, opened 2022-11-10
Ensemble Model (Closes #431)

### LINKED ISSUE #431: Ensemble Model
**Requirement Text**
Ability to handle Ensemble of multiple models

**Suggested Solution**
Ensemble Model class that will run all and average

**DoD**
 - [x] Implement feature
 - [x] Add to example 
 - [x] Add tests 
 - [x] update documentation
 - [x] Add to change notes for next release

teubert:
Creates a new "Ensemble model" that combines one or more models. Each model is executed individually and the results are aggregated using the configurable aggregation method. 

This change also creates an example the example has 3 parts:
1. Ensemble model of 3 BatteryCircuit models with different configuration. The third is very wrong (outlier). The result is wildly inaccurate as a result. 
2. The same model with a median aggregation method. The median aggregation will eliminate the outlier, leading to good performance. 
3. A new ensemble model with members: BatteryCircuit and BatteryElectroChem, demonstration heterogeneous ensemble model. The two models have the same outputs but different states (except for one- battery temperature).

### DIFF
--- a/examples/__init__.py
+++ b/examples/__init__.py
@@ -1,4 +1,4 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the
 # National Aeronautics and Space Administration.  All Rights Reserved.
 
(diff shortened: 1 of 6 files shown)


matteocorbetta (approved):
Approved. Will have to introduce more ensemble model features in the future to provide more comprehensive "ensemble model functionalities." Some of those are likely going to be part of prog_algs.

PR #443, opened 2022-12-02
Hotfix v1.4.4

teubert:
Fix issue with Measurement Noise in LazySimResult. When using LazySimResult, noise is not added to the output. Issue first identified here: https://github.com/nasa/prog_algs/issues/220. A test was added to ensure this continues working

Also, updated one of the error messages.


kjjarvis (approved):
Good bugfix for measurement_noise

PR #446, opened 2022-12-06
hotfix v1.4.5: Fix save pt behavior 

teubert:
This PR fixes an issue in DMD where 1e99 was appended to the end of save_pts, breaking this feature. 

The issue was originating from python's way of using references for objects. 1e99 was appended inside simulate_to_threshold, making the same change in kwargs, which effected the behavior in the larger scope of DMD. 

The fix moves away from appending 1e99 in favor of adding an additional check.

Bug was found by @kjjarvis 

Also added missing copyright

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ INSTALL_REQS = [
 
 setup(
     name = 'prog_models',
-    version = '1.4.4', #pkg_resources.require("prog_models")[0].version,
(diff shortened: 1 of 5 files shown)


teubert (commented):
Test looks great- thanks Katy

kjjarvis (approved):
Nice solution to issue with save_pts, good added tests

PR #449, opened 2022-12-10
Direct Model Interface

teubert:
This PR introduces the concept of a "Direct Model". Direct models map from state to time of event directly without having to perform state transition.

Direct models are introduced through a new method time_of_event which returns the time of event for each event as a dictionary. For direct models this function is overwritten.

When this function is not overwritten-simulate to threshold is called to estimate this value using simulation. 

The pr also includes an example, which calls time of event for direct model and non-direct model. This PR also introduces a is_direct_model property which returns true if time_of_event is overwritten. Tests were also added for both a direct model inherited from a state transition model and one that isn't.

In order to support simple direct models, I had to make outputs optional. The tests are now updated accordingly.

### DIFF
--- a/examples/direct_model.py
+++ b/examples/direct_model.py
@@ -33,7 +33,7 @@ def run_example():
             t_impact = -(x['v'] + np.sqrt(x['v']*x['v'] - 2*g*x['x']))/g
             # 0 = v0 - g*t
             t_falling = -x['v']/g
-            return {'impact': t_impact, 'falling': t_falling}
+            return {'falling': t_falling, 'impact': t_impact}
 
     # Note that adding *args and **kwargs is optional.
     # Having these arguments makes the function interchangeable with other models
--- a/examples/direct_model.py
+++ b/examples/direct_model.py
@@ -1,0 +1,57 @@
+# Copyright © 2021 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration.  All Rights Reserved.
+
+"""
+This example demonstrates the Direct Models functionality. Direct models are models that directly estimate time of event from system state, rather than using state transition. This is useful for data-driven models that map from sensor data to time of event, and for physics-based models where state transition differential equations can be solved.
+
+Here a DirectThrownObject model is defined, extending the ThrownObject state transition model from the new_model example. This model is then initialized and used to estimate the time of event from the initial state.
+"""
+
+import time
+import numpy as np
+from prog_models.models import ThrownObject
+
+def run_example():
+    # Here is how estimating time of event works for a timeseries model
+    # TODO(CT): Time
+    m = ThrownObject()
+    x = m.initialize()
+    print(m.__class__.__name__, "(Direct Model)" if m.is_direct_model else "(Timeseries Model)")
+    tic = time.perf_counter()
+    print('Time of event: ', m.time_of_event(x, dt = 0.05))
+    toc = time.perf_counter()
+    print(f'execution: {(toc-tic)*1000:0.4f} milliseconds')
+
+    # Step 1: Define DirectModel
+    # In this case we're extending the ThrownObject model to include the  time_to_event method, defined in DirectModel
+    # In the case of thrown objects, we can solve the differential equation 
+    # to estimate the time at which the events occur.
+    class DirectThrownObject(ThrownObject):
+        def time_of_event(self, x, **kwargs):
+            # calculate time when object hits ground given x['x'] and x['v']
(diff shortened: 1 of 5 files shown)


kjjarvis (approved):
Nice implementation of a direct model

matteocorbetta (approved):
Looks good to me. I'll work on a data-driven example of DirectModel for a more "meaty" case study.

teubert:
Follow-up for everyone (@matteocorbetta @kjjarvis @AliPohya), I added the following documentation to the next version of the progpy documentation.

This was added to the gl
Screen Shot 2022-12-16 at 9 44 15 AM
ossary:

And this was added to the guide under the "building new models" section
Screen Shot 2022-12-16 at 9 51 36 AM

Let me know if you have any thoughts/comments

examples/direct_model.py
```diff
@@ -0,0 +1,60 @@
...
+    tic = time.perf_counter()
+    print('Time of event: ', m.time_of_event(x))
+    toc = time.perf_counter()
+    print(f'execution: {(toc-tic)*1000:0.4f} milliseconds')
+
+    # Notice that execution is MUCH faster for the direct model. 
+    # This is even more pronounced for events that occur later in the simulation.
+
+    # In this case, the DirectThrownObject has a defined next_state and output equation, 
+    # allowing it to be used with a state estimator (e..g, Particle Filter)
```
AliPohya, line 56: e.g. (typo)

examples/direct_model.py
```diff
@@ -0,0 +1,60 @@
...
+    # to estimate the time at which the events occur.
+    class DirectThrownObject(ThrownObject):
+        def time_of_event(self, x, *args, **kwargs):
+            # calculate time when object hits ground given x['x'] and x['v']
+            # 0 = x0 + v0*t - 0.5*g*t^2
+            g = self.parameters['g']
+            t_impact = -(x['v'] + np.sqrt(x['v']*x['v'] - 2*g*x['x']))/g
+            # 0 = v0 - g*t
+            t_falling = -x['v']/g
+            return {'impact': t_impact, 'falling': t_falling}
```
AliPohya, line 36: switch this so the outputs in the examples are in the same sequence (first: falling, then: impact).

src/prog_models/prognostics_model.py
```diff
@@ -598,7 +601,61 @@ def threshold_met(self, x) -> dict:
         return {key: event_state <= 0 \
             for (key, event_state) in self.event_state(x).items()} 
 
-    def simulate_to(self, time : float, future_loading_eqn : Callable, first_output = None, **kwargs) -> namedtuple:
+    @property
+    def is_direct_model(self) -> bool:
+        """
+        If the model is a "direct model" - i.e., a model that directly estimates time of event from system state, rather than using state transition. This is useful for data-driven models that map from sensor data to time of event, and for physics-based models where state transition differential equations can be solved.
+
+        Returns:
```
AliPohya, line 609: Does this docstring show up correctly in the API documentation? (Asking because the "Returns" is not "underlined")

src/prog_models/prognostics_model.py
```diff
@@ -598,7 +601,61 @@ def threshold_met(self, x) -> dict:
...
+            bool: if the model is a direct model
+        """
+        return type(self).time_of_event != PrognosticsModel.time_of_event
+
+    def time_of_event(self, x, future_loading_eqn = lambda t,x=None: {}, **kwargs) -> dict:
```
AliPohya, line 614: This replacement/override is excellent. I was wondering if some users might want to have both, a direct functionality and a state-transition based functionality, at the same time.

AliPohya (approved):
Looks great! Minor comments on my side, only.

matteocorbetta:
> Follow-up for everyone (@matteocorbetta @kjjarvis @AliPohya), I added the following documentation to the next version of the progpy documentation.
> 
> This was added to the glossary: Screen Shot 2022-12-16 at 9 44 15 AM
> 
> And this was added to the guide under the "building new models" section Screen Shot 2022-12-16 at 9 51 36 AM
> 
> Let me know if you have any thoughts/comments

Chris, I like the descriptions but I'd just add this few words about the input to those models, if you agree to that (see bold text). Just to make it more friendly to data-driven nomenclature.

In the glossary:
A model where the time of event is directly estimated from the current state **or input variables (e.g., sensor data)**, instead of predicted through simulation to threshold.

In the building new model section:
Direct-prediction models
Direct-prediction models are models that estimate time of event directly from the current state **or input variables (e.g., sensor data)**, instead of being predicted through state transition.

up to you tho!

PR #456, opened 2022-12-13
Added future loading classes

teubert:
Adds classes for complex future loading behavior (e.g., noise or moving average). 

Updated futureloading example


portia19 (approved):
For plots of inputs and event_states, what's the respective xlabel? Is it time? It might help to add that in the plot.

teubert:
> For plots of inputs and event_states, what's the respective xlabel? Is it time? It might help to add that in the plot.

That's right- it's time in seconds. I added the xlabel to all the plots

PR #462, opened 2022-12-20
PCE Surrogate Model

teubert:
This is the first draft PCE Surrogate Direct Model. This PR implements the PCE example created by @AliPohya in ProgPy. The result is a surrogate direct model that can be generated from a state transition model. 

This PR includes the new PolynomialChaosExpansion class and an example using this to represent the behavior of an electrochemistry battery. 

To get this to work, a few other changes were required:
1. Removing the requirement for there to be at least one state
2. x is by default none for simulate_to*, making the check easier.  
3. Added models.test_models directory for a few simple models that are used in testing. These are very simple well-defined models. This functionality will not be documented in the user-facing documentation.

Implements #459


matteocorbetta:
> 

Hi Ali, Chris, 
I've been looking into this code; can someone tell me what the parameter discretization is doing?
The description in PCE.from_model says "Number of points to discretize each input."
I thought it defines how many points to use to define the input space, but larger numbers produce worse results.
So I guess I'm missing what it is? 
thanks!

teubert:
> > 
> 
> 
> 
> Hi Ali, Chris, 
> 
> I've been looking into this code; can someone tell me what the parameter discretization is doing?
> 
> The description in PCE.from_model says "Number of points to discretize each input."
> 
> I thought it defines how many points to use to define the input space, but larger numbers produce worse results.
> 
> So I guess I'm missing what it is? 
> 
> thanks!

Good question- youre right that its not clear. We should come up with a description to fix that. 

Basically it's the discretization of future loading for encoding. So if you have a load profile that is 5000 seconds, and a discretization of 6, it will use the load at 0, 1000, 2000, ..., 5000 to represent the profile. But a discretization of 12 would use 0, 500, 1000, 1500, ... , 5000. 

The higher the discretization the more complex load profiles can be encoded, but at a cost of complexity. You will likely need more samples to accurately train a model with higher discretization, requiring more training time.

The input to the PCE is the future load encoded using this discretization, and the output is time of event. 

Does that make sense?

teubert:
Follow-up to last comment. I guess we should take a step back and think about how we're passing in the information about how we're encoding future load. 

Right now it's: max_time, discretization 

But we could possibly do just an list of times. Maybe that would be better. What do you think?

matteocorbetta:
> > Hi Ali, Chris,
> > I've been looking into this code; can someone tell me what the parameter discretization is doing?
> > The description in PCE.from_model says "Number of points to discretize each input."
> > I thought it defines how many points to use to define the input space, but larger numbers produce worse results.
> > So I guess I'm missing what it is?
> > thanks!
> 
> Good question- youre right that its not clear. We should come up with a description to fix that.
> 
> Basically it's the discretization of future loading for encoding. So if you have a load profile that is 5000 seconds, and a discretization of 6, it will use the load at 0, 1000, 2000, ..., 5000 to represent the profile. But a discretization of 12 would use 0, 500, 1000, 1500, ... , 5000.
> 
> The higher the discretization the more complex load profiles can be encoded, but at a cost of complexity. You will likely need more samples to accurately train a model with higher discretization, requiring more training time.
> 
> The input to the PCE is the future load encoded using this discretization, and the output is time of event.
> 
> Does that make sense?

Now it's much clearer, thanks. I am unsure about whether changing the way it is defined. I don't know enough to make a good guess. I need to learn a little more. We can also catch-up one of these days; I think talking to you guys will be helpful for me.

examples/chaos.py
```diff
@@ -0,0 +1,95 @@
...
+    # First let's define the base model that we're creating a surrogate for.
+    m = BatteryElectroChemEOD(process_noise = 0) 
+    x0 = m.initialize()  # Initial State
+    
+    # Step 2: Build surrogate
+    # Next we build the surrogate model from the base model
+    # To build the model we pass in the distributions of possible values for each input.
+    # We also provide the max_time. This is the maximum time that the surrogate will be used for.
+    # We dont expect any battery to last more than 4000 seconds given the high discharge curves we're passing in.
+    m_surrogate = PCE.from_model(m,
```
AliPohya, line 48: Just checking: Did you mean to not include a PCE order that is specified by the user? Including it might be too much for an introductory example - I'm not sure.

src/prog_models/data_models/__init__.py
```diff
@@ -4,8 +4,10 @@
 from .data_model import DataModel
 from .dmd import DMDModel
 from .lstm_model import LSTMStateTransitionModel
+from .pce import PolynomialChaosExpansion, PCE
 
 SURROAGATE_METHOD_LOOKUP = {
```
AliPohya, line 9: (not really important but here's a typo: SurroAgate)

src/prog_models/data_models/__init__.py
```diff
@@ -4,8 +4,10 @@
 from .data_model import DataModel
 from .dmd import DMDModel
 from .lstm_model import LSTMStateTransitionModel
+from .pce import PolynomialChaosExpansion, PCE
```
AliPohya, line 7: aren't PCE and PolynomialChaosExpansion the same object?

src/prog_models/data_models/pce.py
```diff
@@ -0,0 +1,187 @@
...
+
+    Generally this is used as a :term:`surrogate model` for a :term:`model` that is too expensive to simulate. This is done using the :meth:`PolynomialChaosExpansion.from_model` method. The model is used to generate data, which is then used to train the polynomial chaos expansion. The polynomial chaos expansion can then be used to predict the time of event for future inputs.
+
+    Args:
+        models (List[chaospy.Poly]):
+            Polynomial chaos expansion models (one for each event)
+        times (list[float]):
+            List of times to use for the polynomial chaos expansion
+        input_keys (list[str]):
+            List of input keys for the inputs
```
AliPohya, line 25: Does it make sense to give an example?

src/prog_models/data_models/pce.py
```diff
@@ -0,0 +1,187 @@
...
+    @classmethod
+    def from_data(cls, times, inputs, time_of_event, input_keys, **kwargs):
+        """
+        Create a PolynomialChaosExpansion from data.
+
+        Args:
+            times (list[float]):
+                list of times data for use in data. Each element is the time such that inputs[i] is the inputs at time[i]
+            inputs (np.array): 
+                list of :term:`input` data for use in data. Each  eelement is the inputs for a single run of size (n_samples, n_inputs*n_times)
```
AliPohya, line 62: typo

src/prog_models/data_models/pce.py
```diff
@@ -0,0 +1,187 @@
...
+        if len(times) == 0:
+            raise ValueError('Times must include at least one time')
+        if len(inputs) == 0:
+            raise ValueError('Inputs must include at least one run')
+        if len(time_of_event) != len(inputs):
+            raise ValueError('There must be the same number of runs for inputs and time of event')
+
+        n_events = len(time_of_event[0])
+        if n_events == 0:
+            raise ValueError('There must be at least one event to train an PCE model')
```
AliPohya, line 103: typo

src/prog_models/data_models/pce.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Train
+        inputs = inputs.T
+        expansion = cp.generate_expansion(order=params['order'], dist=params['J']) # Order=2 is the only hyperparameter
```
AliPohya, line 107: comment is an old fragment (potentially confusing)

tests/test_surrogates.py
```diff
@@ -528,6 +531,57 @@ def load_eqn(t = None, x = None):
...
+    def _pce_tests(self, m):
+        x0 = m.initialize()
+        input_dists = {'u1': cp.Uniform(0.5, 2), 'u2': cp.Uniform(0.5, 2)}
+        # This is to handle cases where there are <2 inputs
+        input_dists = {key: input_dists[key] for key in m.inputs}
+        pce = PCE.from_model(m, x0, input_dists, times = [i*10 for i in range(5)], N = 250)
+        pce_result = pce.time_of_event(x0, lambda t, x=None: pce.InputContainer({'u1': 1, 'u2': 0.75}))
+        gt_result = m.time_of_event(x0, lambda t, x=None: m.InputContainer({'u1': 1, 'u2': 0.75}))
+        for event in m.events:
+            self.assertAlmostEqual(pce_result[event], gt_result[event], delta=1)
```
AliPohya, line 553: Wow, this is nice!

AliPohya (approved):
I really like it!

PR #477, opened 2023-01-09
Composite Model

teubert:
Creates CompositeModel class, a class used to create a model of a system of inter-related systems, where the outputs or states of one are the input to another. The resulting composite model is a fully-functional prognostic model.

Also includes an example of its use, and tests. Adds the following additional features:
* OneInputOneOutputNoEventLM test model
* Container.get which works the same as dict.get


kjjarvis (approved):
Very nice implementation of composite models - easy to use and clear implementation

PR #480, opened 2023-01-13
Feature/traj generation

kjjarvis:
This model describes the dynamics of a small UAV and generates predictions of a flyable trajectory given a set of waypoints. 

This work was previously developed by @matteocorbetta and @portia19, and model development is described in Corbetta et al. 2019, "Real-time UAV trajectory prediction for safely monitoring in low-altitude airspace". 

PR includes all necessary functionality for:
- Vehicle model (including controller, multiple vehicle types, etc.) 
- Trajectory generator (NURBS algorithm) 

Future work described here: #481


src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+from prog_models.exceptions import ProgModelInputException
+
+class UAVGen(PrognosticsModel):
+    """
+    Vectorized prognostics :term:`model` to generate a predicted trajectory for a UAV. 
+    
+    Model generates cartesian positions and velocities, pitch, roll, and yaw, and angular velocities throughout time to satisfy some user-define waypoints. 
+
+    :term:`Events<event>`: (1)
+        TrajectoryComplete: percentage of waypoints remaining for trajectory to pass through
```
teubert, line 24: The event would actually be completion of all waypoints, then the event state is the percentage of waypoints completed.

```suggestion
        TrajectoryComplete: All waypoints are completed
```

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+    
+    :term:`Inputs/Loading<input>`: ()
```
teubert, line 26:
```suggestion
    :term:`Inputs/Loading<input>`: (0)
```

The way we define inputs for ProgPy, for now this model has 0 inputs. Instead everything is configuration parameters. 

This will change with the "controller as future loading" change we discussed

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+          additional column for ETAs may be included (labeled 'time_unix). Note that while
+          'flight_file' is optional, either 'flight_file' or 'flight_plan' must be specified.
+        flight_plan : Optional, dict[str, numpy array]
+          Dictionary to specify waypoint information. Necessary keys must include the following
+          with units specified: latitude ('lat_deg' or 'lat_rad'), longitude 
+          (labeled 'lon_deg' or 'lon_rad'), and altitude (labeled 'alt_ft' or 'alt_m'). An 
+          additional key for ETAs may be included (labeled 'time_unix). Each key must correspond
+          to a numpy array of values. Note that while 'flight_plan' is optional, either 
+          'flight_file' or 'flight_plan' must be specified.
+        flight_name : Optional, str
```
teubert, line 91: What do we mean by name to identify flight plan? I see that it's passed into trajectory and route building, but what does it add?

teubert: same with aircraft_name

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+          Optional string to identify flight plan.
+        aircraft_name : Optional, str
+          Optional string to identify aircraft. 
+        dt : float
```
teubert, line 95: These and all the params below are optional as well, right?

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+        | phi: pitch
+        | theta: roll
+        | psi: yaw
+        | vx: velocity 
+        | vy: velocity 
+        | vz: velocity 
+        | p: angular velocity 
+        | q: angular velocity 
+        | r: angular velocity 
+        | t: time
```
teubert, line 62:
```suggestion
```

Time isn't an output, right? It's not in model.outputs, below

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+    :term:`Inputs/Loading<input>`: ()
+        | User-defined inputs: waypoints plus ETAs or speeds
+        | Model-defined inputs: 
+            | T: thrust
+            | mx: moment in x 
+            | my: moment in y
+            | mz: moment in z
+
+    :term:`States<state>`: (13)
+        | x: cartesian position
```
teubert, line 35: Is there a way to specify the direction of x, y, z?

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+        flightplan = trajectory.load.convert_dict_inputs(self.parameters['flight_plan'])
+        lat = flightplan['lat_rad']
+        lon = flightplan['lon_rad']
+        alt = flightplan['alt_m']
+        tstamps = flightplan['timestamp']
+      elif self.parameters['flight_file'] != None and self.parameters['flight_plan'] == None:
+        flightplan = trajectory.load.get_flightplan(fname=self.parameters['flight_file'])
+        lat, lon, alt, tstamps = flightplan['lat'], flightplan['lon'], flightplan['alt'], flightplan['timestamp']
+      elif self.parameters['flight_file'] != None and self.parameters['flight_plan'] != None:
+        raise ProgModelInputException("Too many flight plan inputs - please input either flight_plan dictionary or flight_file.")
```
teubert, line 192:
```suggestion
        raise ProgModelInputException("Too many flight plan inputs - please input either flight_plan dictionary or flight_file, not both.")
```

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+      else:
+        raise ProgModelInputException("No flight plan information supplied.")
```
teubert, line 194:
```suggestion
        raise ProgModelInputException("No flight plan information supplied. Please provide flight_plan or flight_file")
```

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+
+      super().__init__(**kwargs)
+      
+      if self.parameters['flight_plan'] is not None and self.parameters['flight_file'] == None:
+        # Check for appropriate input:
+        if not isinstance(self.parameters['flight_plan'], dict):
+          raise ProgModelInputException("'flight_plan' must be a dictionary. Type {} was given".format(type(self.parameters['flight_plan'])))
+        for key in self.parameters['flight_plan'].keys():
+          if not isinstance(self.parameters['flight_plan'][key], np.ndarray):
+            raise ProgModelInputException("'flight_plan' entries must be numpy arrays specifying waypoint information. Type {} was given".format(type(self.parameters['flight_plan'])))
```
teubert, line 181:
```suggestion
        for flight_plan_element in self.parameters['flight_plan'].values():
          if not isinstance(flight_plan_element, np.ndarray):
            raise ProgModelInputException("'flight_plan' entries must be numpy arrays specifying waypoint information. Type {} was given".format(type(flight_plan_element)))
```

teubert (commented):
Submitting partial review. I'm still working on it. 

Great work! It looks really great and I'm so excited that you have it working. I made a few suggestions. Let me know if you have any questions. Otherwise, I'll continue reviewing tomorrow.

Have a great long weekend!

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+          'vy': self.ref_traj.velocity[0,1],
+          'vz': self.ref_traj.velocity[0,2],
+          'p': self.ref_traj.angular_velocity[0,0],
+          'q': self.ref_traj.angular_velocity[0,1],
+          'r': self.ref_traj.angular_velocity[0,2],
+          't': 0
+          })
+    
+    def dx(self, x : dict, u : dict):
+
```
teubert, line 269: This method is very well written and commented. Easy to understand

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+        ]))
+    
+    def event_state(self, x : dict) -> dict:
+        # Extract next waypoint information 
+        num_wypts = len(self.parameters['waypoints']['waypoints_time']) - 1 # Don't include initial waypoint
+        index_next = self.parameters['waypoints']['next_waypoint']
+
+        # Check if at intial waypoint. If so, event_state is 1
+        if index_next == 0:
+            self.parameters['waypoints']['next_waypoint'] += 1
```
teubert, line 357:
```suggestion
            self.parameters['waypoints']['next_waypoint'] = 1
```

Microoptimization

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+            np.atleast_1d(dxdt[9]),
+            np.atleast_1d(dxdt[10]),
+            np.atleast_1d(dxdt[11]),
+            np.atleast_1d(dxdt[12])
+        ]))
+    
+    def event_state(self, x : dict) -> dict:
+        # Extract next waypoint information 
+        num_wypts = len(self.parameters['waypoints']['waypoints_time']) - 1 # Don't include initial waypoint
+        index_next = self.parameters['waypoints']['next_waypoint']
```
teubert, line 353: The model itself should be stateless. Some algorithms might require that you run multiple states through the model (e.g., when building a surrogate model). This will break down if you save state in the parameters. In this case, the next_waypoint is part of the state of the system. 

next_waypoint should be part of the state vector and updated in next_state, or it should be calculated each time. 

Since this is a bigger change, I'm ok with leaving it as is (with a note about this limitation) if you add this to the github issue for the next release.

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+            'theta': x['theta'],
+            'psi': x['psi'],
+            'vx': x['vx'],
+            'vy': x['vy'],
+            'vz': x['vz'],
+            'p': x['p'],
+            'q': x['q'],
+            'r': x['r'],
+            't': x['t']
+            })
```
teubert, line 424:
```suggestion
        return self.OutputContainer(x.matrix)
```

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+                return {
+                    'TrajectoryComplete': True
+                }
+            else: 
+                return {
+                    'TrajectoryComplete': False
+                }
+        else: 
+            # Trajectory has passed acceptable bounds of final ETA - simulation terminated
+            warn("Trajectory simulation extends beyond the final ETA. Either the final waypoint was not reached in the alotted time (and the simulation was terminated), or simulation was run for longer than the trajectory length.")
```
teubert, line 447: What if the final waypoint hasn't been met yet, but we're running behind? Is that a case you would want to support

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+                u = self.vehicle_model.control_scheduled(x_temp - ref_now) 
+                u[0]      += self.vehicle_model.steadystate_input
+                u[0]       = min(max([0., u[0]]), self.vehicle_model.dynamics['max_thrust'])
+                return self.InputContainer({'T': u[0], 'mx': u[1], 'my': u[2], 'mz': u[3]})
+
+        # Simulate to threshold 
+        results = super().simulate_to_threshold(future_loading_new,first_output, threshold_keys, **kwargs)
+        return results 
+
+    def visualize_traj(self, pred):
```
teubert, line 493: This could use some doc comments (triple quotes) after the function definition describing the function. These will then show up on the website documentation

teubert (commented):
A few more comments. I'll look more this afternoon

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,556 @@
...
+from prog_models.models.uav_model.utilities import geometry
+
+import numpy as np
+import datetime
+from copy import deepcopy
+import prog_models.models.uav_model.utilities.geometry as geom
+from warnings import warn
+from prog_models.exceptions import ProgModelInputException
+
+class UAVGen(PrognosticsModel):
```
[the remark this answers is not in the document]
matteocorbetta: That's a good question and we've been talking about it.
Probably UAVFlight would be the best in this situation. However, we've also talked about increasing the ``modularity'' of this model, but generating a controller that then is assigned to the UAV model. Also, trajectory generation is independent of the UAV model (up to minor things like: what type of UAV it is (rotorcraft vs. fixed wind vs. something else, maximum speed/acceleration possible, and stuff like that). 
So maybe we should re-factor this to make it indeed modular? don't know if we want to leave it for the future.

examples/uav_dynamics_model.py
```diff
@@ -0,0 +1,123 @@
...
+        'vehicle_model': 'tarot18', # Define vehicle
+    }
+
+    # Create a model object, define noise
+    uav_1 = UAVGen(**params_1)
+    uav_1.parameters['process_noise'] = 0
+
+    # Define future loading function to return empty InputContainer, since there is no user-defined loading in trajectory generation 
+    def future_loading_1(t, x=None):
+        return uav_1.InputContainer({})
```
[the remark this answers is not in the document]
kjjarvis: Oh great! I think this is a great change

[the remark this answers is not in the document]
kjjarvis: Actually, this doesn't work. I get an error, "simulate_to_threshold() missing 1 required positional argument: 'future_loading_eqn' ". I checked a quick simulation with the ThrownObject, and leaving out future_loading works fine, so I'm wondering if it doesn't work here because we have inputs under the hood. If so, another argument to make some changes for modularity soon. I'll keep debugging

src/prog_models/models/uav_model/uav_model.py
kjjarvis: Good point. It doesn't add anything in terms of the actual trajectory generation. The original code had capability to save results to a file, and I believe the name(s) were used for specifying the file name. I didn't bring that capability over, so it's not used here. @portia19 or @matteocorbetta, can you confirm I'm understanding this correctly? Are we okay with not including the saving capabilities (users could use our ProgPy serialization techniques instead) and getting rid of aircraft_name and flight_name?

src/prog_models/models/uav_model/uav_model.py
kjjarvis: Most are, good catch. The speeds are funny because they are either *not* an option if ETAs are defined (any input values are ignored and warning is printed) or *required* if no ETAs are defined. Do you think these should be listed as Optional?

src/prog_models/models/uav_model/uav_model.py
kjjarvis: @teubert I found a mistake in my output function, since it includes time and we noted above that this isn't an output. I'll change this to "return self.OutputContainer(x.matrix[0:-1]) " . It seems to work - is that okay?

src/prog_models/models/uav_model/uav_model.py
kjjarvis: It will continue simulating for a buffer time defined by the parameter "final_time_buffer_sec" past the final ETA time. Originally I implemented it this way because I was thinking that if the final waypoint wasn't met, the simulation would go on forever. But now I'm realizing that the default horizon would eventually stop this. However, once we pass the final ETA, we won't have any new input information, so the results quickly level out to whatever output corresponds to that input. Does that make sense? I'm happy to leave it as is or get rid of this and let it run to horizon

src/prog_models/models/uav_model/uav_model.py
kjjarvis: @teubert I'd love to talk with you more about this. For now, I've added it to the future work issue but hopefully we can get it done before the PR is finalized

matteocorbetta:
I noticed that, generally in prognostic_model, the method simulate_to_threshold initializes u before initializing x, such that when future_loading_eqn is called the first time x=None.
for this specific case, initializing x first would be beneficial, as we would skip the if statement in future_loading_eqn all together. Would be possible to initialize x before u in simulate_to_threshold? what's the con about it?

I'm having a deja-vu (I believe we already talked about it but not sure what the problem would be).

src/prog_models/models/uav_model/uav_model.py
teubert: sounds great

src/prog_models/models/uav_model/vehicles/AircraftModels.py
```diff
@@ -0,0 +1,268 @@
...
+    # Default control parameters (if not provided)
+    if params['Q'] is None:
+        params['Q'] = np.diag([1000, 1000, 25000,  # x, y, z
+                               100.0, 100.0, 100.0,  # phi, theta, psi,
+                               1000, 1000, 5000,  # vx, vy, vz,
+                               1000, 1000, 1000])  # p, q, r
+    if params['R'] is None:         params['R']     = np.diag([500, 4000, 4000, 4000])   # T, Mu, Mv, Mw
+    if params['qi'] is None:        params['qi']    = np.array([100, 100, 1000])  # Integral error weights in position x, y, z
+    if params['i_lag'] is None:     params['i_lag'] = 100                        # integral lag: how far back to use (in data points, so 1 point = 1 dt) for integral error
+
```
matteocorbetta, line 39: here I cleaned up some controller parameter's definition which have been there for a while (my fault). The trajectories now look a little cleaner as well.

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,567 @@
...
+          if self.parameters['cruise_speed'] is None or self.parameters['ascent_speed'] is None or self.parameters['descent_speed'] is None or self.parameters['landing_speed'] is None:
+              raise ProgModelInputException("ETA or speeds must be provided. If ETAs are not defined, desired speed (cruise, ascent, descent, landing) must be provided.")  
+          route_ = route.build(name=self.parameters['flight_name'], lat=lat, lon=lon, alt=alt, departure_time=tstamps[0],
+                                parameters = self.parameters)
+
+      # Save final waypoint information for threshold_met and event_state 
+      coord_end = geom.Coord(route_.lat[0], route_.lon[0], route_.alt[0])
+      self.coord_transform = deepcopy(coord_end)
+      wypt_time_unix = [datetime.datetime.timestamp(route_.eta[iter]) - datetime.datetime.timestamp(route_.eta[0]) for iter in range(len(route_.eta))]
+      wypt_x, wypt_y, wypt_z = coord_end.geodetic2enu(route_.lat, route_.lon, route_.alt) # computing way-points in ENU in one shot
```
matteocorbetta, line 252: here there was a for loop I substituted (functions in geom.Coord work with vectors).

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,567 @@
...
+        dxdt[10] = ((Izz - Ixx) * p * r + tq * self.vehicle_model.geom['arm_length']) / Iyy
+        dxdt[11] = ((Ixx - Iyy) * p * q + tr *        1               ) / Izz
+        dxdt[12] = 1
+
+        # Set vehicle state:
+        state_temp = np.array([x[iter] for iter in x.keys()])
+        self.vehicle_model.set_state(state=state_temp + dxdt*self.parameters['dt'])
+        
+        # I'd suggest a more compact way of generating the StateContainer
+        return self.StateContainer(np.array([np.atleast_1d(item) for item in dxdt]))
```
matteocorbetta, line 361: in-line for loop to generated state container. it's just more compact.

examples/uav_dynamics_model.py
```diff
@@ -0,0 +1,123 @@
...
+    # Define model parameters:
+    params_1 = {
+        'flight_plan': waypoints, # Specify waypoints 
+        'dt': 0.1, # Define time step for model generation; this value is also used for simulate_to functionality 
+        'vehicle_model': 'tarot18', # Define vehicle
+    }
+
+    # Create a model object, define noise
+    uav_1 = UAVGen(**params_1)
+    uav_1.parameters['process_noise'] = 0
```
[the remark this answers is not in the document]
teubert: if you're adding to params_1, then you can delete this line
```suggestion
```

src/prog_models/models/uav_model/uav_model.py
teubert: OK- I see why you implemented it that way it is. If you're simulating to threshold wouldn't it stop anyways at the last waypoint, so it wouldn't simulate past the last input information. 

I guess that doesn't help for simulate to. I'm hoping you can remove this when we move to the modular, since input would be provided by the future loading equation instead. I don't like the idea of having two horizons

src/prog_models/models/uav_model/uav_model.py
teubert: I think this one was missed. I still see output here, but it's not in outputs on line 154

src/prog_models/models/uav_model/uav_model.py
teubert: Hmmm... good question. I would say, lets leave the optional flag off of those

src/prog_models/models/uav_model/uav_model.py
```diff
@@ -0,0 +1,577 @@
...
+          kg, payload mass
+
+    References 
+    ----------
+        References 
+    -------------
+     .. [0] M. Corbetta et al., "Real-time UAV trajectory prediction for safely monitoring in low-altitude airspace," AIAA Aviation 2019 Forum,  2019. https://arc.aiaa.org/doi/pdf/10.2514/6.2019-3514
+    """
+    events = ['TrajectoryComplete']
+    inputs = ['T','mx','my','mz']
```
teubert, line 152:
```suggestion
    inputs = []
```

Since 'T','mx','my','mz' are calculated by the controller, which is internal to the model, I think these should not be listed as inputs. Do I understand that correctly?

teubert (commented):
A few mostly cosmetic suggestions. I want to run a few more tests tomorrow, then I think I'll be done with the review

src/prog_models/models/uav_model/uav_model.py
kjjarvis: I think this is an issue with our current implementation, where the "inputs" are defined in the internal "future_loading" function. They still act as inputs in the normal sense of u in the overall ProgPy framework, even though they aren't user defined. If I change the code as suggested here, I get an error, since u is still used throughout (for example, in the dx calculation on line 309). The more we review, the more I think it's time to switch to a more modular version ...

src/prog_models/models/uav_model/uav_model.py
kjjarvis: I've removed flight_name and aircraft_name in the new PR with modularity changes, so this will be adjusted soon

PR #485, opened 2023-01-26
Added is_continuous/discrete

teubert:
Added is_continuous and is_discrete properties


matteocorbetta (approved):
Great! I didn't know of this syntax ``type(self).method''
Tried it out. looks good!.

Issue #479, opened 2023-01-10
Composite: Performance metrics


teubert:
Extend composite models to support performance metrics

PR #486, opened 2023-01-27
Composite Performance Metrics

teubert:
Added Performance Metrics to Composite Model (see #479). Previous implementation didn't include logic to handle models with performance metrics. Added the feature and tests. 

Tests include case where there are no performance metrics (default), and a case with performance metrics in each of the models.


src/prog_models/models/test_models/linear_models.py
```diff
@@ -67,6 +67,16 @@ class OneInputOneOutputNoEventLM(LinearModel):
     }
 
 
+class OneInputOneOutputNoEventLMPM(OneInputOneOutputNoEventLM):
+    """
+    Same as OneInputOneOutputNoEventLM, but with performance metrics defined as a function. Has a single performance metric that is always the state, plus 1
+    """
+    performance_metric_keys = ['x1+1']
```
kjjarvis, line 74: I find this notation to be slightly confusing/hard to read when we implement it in the tests below (e.g. with syntax like 'OneInputOneOutputNoEventLMPM.x1+1'). What if we changed it to something like, 'x1plus1' ? Open to suggestions

kjjarvis (approved):
Clear implementation of performance_metrics for the composite model and good added testing

src/prog_models/composite_model.py
```diff
@@ -206,6 +208,20 @@ def output(self, x):
                 z[name + '.' + key] = value
         return self.OutputContainer(z)
 
+    def performance_metrics(self, x):
+        pf = {}
```
aqitya, line 212: Maybe consider being more descriptive with the variable 'pf' here.

src/prog_models/composite_model.py
```diff
@@ -206,6 +208,20 @@ def output(self, x):
...
+        for (name, m) in self.parameters['models']:
+            # Prepare state
+            x_i = m.StateContainer({key: x[name + '.' + key] for key in m.states})
+
+            # Get outputs
+            pf_i = m.performance_metrics(x_i)
```
aqitya, line 218: Same thing here; keeping the "_i" part is reasonable.

aqitya (approved):
Thank you for having me look over the "reviewer" side of a pull request!

Issue #478, opened 2023-01-10
propeller load model


teubert:
Suggestion from @kjjarvis : Move the propeller load model from the composite example into its own separate model.  

Question to answer: Should propeller load be a model, or part of the future_load eqn.

Original quote:

       Just a thought, I'm not sure if it's a good idea, but we could add this as a simple model in ProgPy (it's already in the powertrain model, right? But we could add it as a stand alone). Then users would have it to implement with the other models in other ways, and it would clean up this example a bit and allow us to highlight the composite model even more. I know that Matteo has been working on a propeller model, so we could also leave this as is and add that model in later when it's ready. Just a thought to consider.

_Originally posted by @kjjarvis in https://github.com/nasa/prog_models/pull/477#discussion_r1066198015_

PR #488, opened 2023-01-28
[Simple] Move Propeller Load Model

teubert:
Very Simple PR. I just moved the propeller load model from the composite example to the prog_models.models package (Implements ticket #478, idea by kjjarvis)


mstraut (approved):
I wanted to note that it failed the test for 3.7, I don't see anything obviously wrong. My assumption is that it is some code that is version specific to 3.8.

PR #491, opened 2023-02-07
changed some code for clarification and added comments

mstraut:


teubert (changes requested):
The target branch for the PR should be dev, not master.

Merging into master is a release, which we only do if something is broken.

PR #494, opened 2023-02-13
Fixed import and moved setter methods to correct location

aqitya:
Moved setter method in the linear_model.py file for parameter E to make code more clear.

### DIFF
--- a/src/prog_models/linear_model.py
+++ b/src/prog_models/linear_model.py
@@ -4,8 +4,7 @@
 from abc import ABC, abstractmethod
 import numpy as np
 
-from . import PrognosticsModel
-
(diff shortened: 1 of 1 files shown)


kjjarvis:
@aqitya this looks great, nice changes. May I suggest that we change the PR title to something like, "linear_model edits for readability", and then include your current title in the comments? What you have is a great description, but it may be easier for traceability if we have "linear_model" in the title of the PR (and I'm sorry I didn't think suggest that before when we did this together!)

kjjarvis (approved):
Nice changes for readability of linear_model

teubert (approved):

Issue #493, opened 2023-02-10
__sizeof__


teubert:
**Requirement Text**
Implement __sizeof__ method for parameters, models, and containers

**Background Information**
__sizeof__ determines the amount of size an object takes

**Suggested Solution**
model: size of object plus size of parameters

minimum test - make sure size of model is greater than size of parameters, and adding new parameter increases size. 

**DoD**
 - [ ] Implement feature
 - [ ] Add tests

Issue #482, opened 2023-01-26
Rework Data Storage


teubert:
**Requirement Text**
Consider alternate solutions for data storage and implement changes where appropriate. Should consider:
* ease of use
* keeping context with data
* use in vectorized applications 
* backwards compatibility 
* future extendability (e.g., windowed state containers)
* efficiency in creation and accessing 
* ability to convert to TF tensor (request from partner)

**Background Information**
There are three related data structures used in ProgPy:
*Containers - dict-like wrappers around a numpy column vector
SimResult- essentially a list of *containers, representing the result of a simulation 
Predicton (from prog_algs)- result of prediction. Essentially a simresult but each element is UncertainData. This may not change as part of the PR

**Suggested Solution**
Two solutions are considered:
* replace simresult with pandas data frame and *container with pandas series (or data frame when vectorized. Idea by @mstraut 
* replace internal representation on simresult and prediction with pandas data frame, numpy array, or tensor flow tensor

**note**
This change may also help simplify plotting features

**DoD**
 - [ ] complete study of options, develop proposal 
 - [ ] Implement change
 - [ ] quantify impact to efficiency
 - [ ] Update tests to include new abilities (e.g., direct access to dataframe and conversion to TF tensor (see https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor)

teubert:
Decided to use Pandas- partially implemented in v1.5, issue moved to new repo (progpy) https://github.com/nasa/progpy/issues/12

PR #496, opened 2023-02-17
Add ability to estimate the size of a model 

teubert:
This is useful for benchmarking and requested by the TechEdSat effort. Now sys.getsizeof works properly for PrognosticsModels and Parameters. 

I added two tests that show that adding to the model increases the size by the expected amount. This is challenging because the size of overhead is python version and computer-dependent. To handle this the tests pass so long as the overhead is "reasonable".

This change also includes changing the default in parameters to "not copy" this allows python's optimizations to stay in place. I originally changed this to allow for test of repeated objects in parameters, but it is also a good change generally. 

This implements the first part of #493. __sizeof__ wasn't implemented for the containers (SimResult, SimResults, and *Containers) because they will soon be replaced as part of #482. Note: #493 is to be kept open for this reason.

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -17,6 +17,7 @@ from .utils import ProgressBar
 from .utils.containers import DictLikeMatrixWrapper
 from .utils.parameters import PrognosticsModelParameters
 from .utils.serialization import *
+from .utils.size import getsizeof
 
 
 class PrognosticsModel(ABC):
@@ -1138,6 +1139,9 @@ class PrognosticsModel(ABC):
             saved_outputs, 
             saved_event_states
         )
+
+    def __sizeof__(self):
+        return getsizeof(self)
 
     def calc_error(self, times : List[float], inputs : List[dict], outputs : List[dict], **kwargs) -> float:
         """Calculate Mean Squared Error (MSE) between simulated and observed
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -11,13 +11,13 @@ import numpy as np
 from typing import Callable, Iterable, List
 from warnings import warn
 
-from .exceptions import ProgModelInputException, ProgModelTypeError, ProgModelException, ProgModelStateLimitWarning
-from .sim_result import SimResult, LazySimResult
-from .utils import ProgressBar
-from .utils.containers import DictLikeMatrixWrapper
-from .utils.parameters import PrognosticsModelParameters
-from .utils.serialization import *
-from .utils.size import getsizeof
+from prog_models.exceptions import ProgModelInputException, ProgModelTypeError, ProgModelException, ProgModelStateLimitWarning
+from prog_models.sim_result import SimResult, LazySimResult
+from prog_models.utils import ProgressBar
+from prog_models.utils.containers import DictLikeMatrixWrapper
+from prog_models.utils.parameters import PrognosticsModelParameters
+from prog_models.utils.serialization import *
+from prog_models.utils.size import getsizeof
 
 
 class PrognosticsModel(ABC):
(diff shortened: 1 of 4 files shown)


src/prog_models/prognostics_model.py
```diff
@@ -17,6 +17,7 @@
 from .utils.containers import DictLikeMatrixWrapper
 from .utils.parameters import PrognosticsModelParameters
 from .utils.serialization import *
+from .utils.size import getsizeof
```
aqitya, line 20: I thought it was decided to not use relative path imports anymore.

teubert: It was, good catch!

mstraut (approved):
Right now the file is consistent with its formatting of imports. My suggestion is that the decision to  change them is something that can be done in another PR.

teubert:
> Right now the file is consistent with its formatting of imports. My suggestion is that the decision to change them is something that can be done in another PR.

Your comment makes sense, but I think we want to start changing the import format as we touch files for now. We already have a mixture of styles, and this way we continue moving forward. 

I made the change and also added a few more tests that I thought of. I'll re-request review.

mstraut (approved):
the updates to the imports look good.

tests/test_base_models.py
```diff
@@ -154,6 +154,71 @@ def next_state(self, x, u, dt):
...
+
+        # Size difference should be slightly more than the size of key & value
+        # The difference is overhead for the dict
+        diff = size3 - size2
+        sum_of_parts = sys.getsizeof('_test_value') + sys.getsizeof(123456789)
+        self.assertLess(sum_of_parts, diff)
+        self.assertLess(diff-sum_of_parts, 100)
+
+        # Add list into parameters
+        m.parameters['test_list'] = [79534, 84392, 93333, -243934, 23233]
```
aqitya, line 193: Small stylistic change to reduce code repeated sections of longer code.

One potential change that you can implement is you can set the list to its own variable and then input that list in to the sys.getsizeof() for the sum_of_parts.

aqitya:
I reran the "test (3.7)" check since it had failed since some of my pushs ended up passing that particular check when I rerun the check, however, it ended up "Cancelling" itself this time around.

Not sure what that means... so I will write down my observations here.

tests/test_base_models.py
```diff
@@ -154,6 +154,71 @@ def next_state(self, x, u, dt):
...
+        # The difference is overhead for the dict
+        # Note- have to use an uncommon number for this to work
+        # This is because of python's memory allocation
+        diff = size2 - size
+        sum_of_parts = sys.getsizeof('test') + sys.getsizeof(8675309)
+        self.assertLess(sum_of_parts, diff)
+        self.assertLess(diff-sum_of_parts, 100)
+
+        # Adding other attributes
+        m._test_value = 123456789
```
aqitya, line 179: '_test_value', is this a typo where it should be 'test_value'?

aqitya: Since we are adding an additional test, maybe the key should be called a bit more descriptive and state "additional_test_value", or something along those lines.

aqitya:
Everything added looks good!

On a side note, I was investigating a way for Python to reallocate memory in an efficient manner so we can run the getsizeof() call in a manner that checks if the size of an object has decreased, which it does not have (a bit of a stretch to look for since if it existed it would have been a widely implemented part of every program but was interesting to see how Python memeory garbage collection works).

If such a way exists, then it would be a good test to add.

tests/test_base_models.py
mstraut: I think its to let people know that the variable is meant to be private. 
https://www.geeksforgeeks.org/private-variables-python/

aqitya: Oh, good point.

I guess I just wrongly assumed it was weird that we were declaring a private variable after declaration and later passing a value into it that I didn't think it was a private variable.

aqitya (approved):

tests/test_base_models.py
teubert: You're right that in practice it's strange to declare a private variable from outside of the scope of the class after its creation. But, I think it's ok for the sake of this test where the variable name doesn't matter

PR #501, opened 2023-03-03
Hotfix v1.4.6

teubert:
This hot fix fixes an issue with the Parameter Estimation feature. Parameter estimation wasn't working correctly. Issue and solution were identified by @kjjarvis


kjjarvis (approved):
Good catch and fix for estimating parameters correctly

PR #509, opened 2023-03-23
fixed example in containers.py

mstraut:
updated an example that didn't work.

### DIFF
--- a/src/prog_models/utils/containers.py
+++ b/src/prog_models/utils/containers.py
@@ -164,7 +164,7 @@ class DictLikeMatrixWrapper():
         example
         -------
         >>> from prog_models.models import DictLikeMatrixWrapper
-        >>> dlmw = DictLikeMatrixWrapper(['a', 'b', 'c'], [[1], [2], [3]])
+        >>> dlmw = ['a', 'b', 'c'], np.array([1, 2, 3], dtype=np.float64)
         >>> 'a' in dlmw  # True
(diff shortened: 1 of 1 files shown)


src/prog_models/utils/containers.py
teubert, line 1: Good catch that this example didn't work. I also noticed the import is wrong and I don't think you meant to convert it to a tuple.

Here's what I think the example should be:

```python
         >>> from prog_models.utils.containers import DictLikeMatrixWrapper
         >>> import numpy as np
         >>> dlmw = DictLikeMatrixWrapper(['a', 'b', 'c'], np.array([1, 2, 3], dtype=np.float64))
         >>> 'a' in dlmw  # True
```

Alternately, we could initialize the DictLikeMatrixWrapper with a dict so we don't need to import numpy:

```python
         >>> from prog_models.utils.containers import DictLikeMatrixWrapper
         >>> dlmw = DictLikeMatrixWrapper(['a', 'b', 'c'], {'a': 1, 'b': 2, 'c': 3})
         >>> 'a' in dlmw  # True
```

What do you think?

teubert (changes requested):

src/prog_models/utils/containers.py
mstraut: Yes, in my modified version I updated it like your second example.

teubert (approved):

PR #512, opened 2023-04-05
Additional Error Methods (Closes #510)

### LINKED ISSUE #510: Calc_error max error
**Requirement Text**
Support calculation of max error of a data with respect to simulated value both directly and in estimate parameters

**Background Information**
Currently, calc_error estimates the MSE, but sometimes max_error is a better metric

**Suggested Solution**
Optional parameter method in calc_error. Default is MSE, but add MAX_E option. Add to estimate_Paramus as well and pass through

**DoD**
 - [ ] Implement feature in calc_error
 - [ ] Implement in estimate_params
 - [ ] Add to example 
 - [ ] Add tests 
 - [ ] update documentation
 - [ ] Add to change notes for next release

teubert:
This idea comes from something Juan was doing with his work. 

This PR adds the ability to select different metric types when calculating error. The default configuration was MSE, but this PR adds RMSE, Max error (what Juan was using), MAE, and MAPE. The change was propagated into estimate_parameters so we can change what method we're using for error population when estimating parameters. 

Partially closes #510 .

It conflicts with Aditya's changes to the tests, so the tests will be added in a separate PR

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -110,31 +110,31 @@ jobs:
     - name: Run copyright check
       run: |
         python scripts/test_copyright.py
-  # coverage:
-  #   timeout-minutes: 30
-  #   runs-on: ubuntu-latest
-  #   strategy:
-  #     matrix:
-  #       python-version: ['3.9']
-  #   steps:
-  #   - uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


mstraut (approved):
It looks good.

PR #514, opened 2023-04-07
Improve Ensemble Tests

teubert:
This PR introduces a set of tests for the ensemble model feature

### DIFF
--- a/src/prog_models/ensemble_model.py
+++ b/src/prog_models/ensemble_model.py
@@ -1,6 +1,7 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the
 # National Aeronautics and Space Administration.  All Rights Reserved.
 
+from collections.abc import Iterable
 import numpy as np
 
 from . import PrognosticsModel
@@ -19,7 +20,7 @@ class EnsembleModel(PrognosticsModel):
     See example :download:`examples.ensemble <../../../../prog_models/examples/ensemble.py>`
 
     Args:
-        models (list): List of models that form the ensemble
+        models (list[PrognosticsModel]): List of at least 2 models that form the ensemble
 
     Keyword Arguments:
         aggregation_method (function): Function that aggregates the outputs of the models in the ensemble. Default is np.mean
@@ -30,6 +31,14 @@ class EnsembleModel(PrognosticsModel):
     }
 
     def __init__(self, models, **kwargs):
+        if not isinstance(models, Iterable):
+            raise TypeError(f'EnsembleModel must be initialized with a list of models, got {type(models)}')
+        if isinstance(models, Iterable) and len(models) < 2:
+            raise ValueError('EnsembleModel requires at least two models')
+        for m in models:
+            if not isinstance(m, PrognosticsModel):
+                raise TypeError(f'EnsembleModel requires all models to be PrognosticsModel instances, was {type(m)}')
+
         inputs = set()
         states = set()
         outputs = set()
@@ -47,7 +56,7 @@ class EnsembleModel(PrognosticsModel):
         super().__init__(**kwargs)
         self.parameters['models'] = models
 
-    def initialize(self, u, z=None):
+    def initialize(self, u=None, z=None):
         xs = [m.initialize(m.InputContainer(u), m.OutputContainer(z) if z is not None else None) for m in self.parameters['models']]
         x0 = {}
         for x in xs:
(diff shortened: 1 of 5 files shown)


tests/__main__.py
```diff
@@ -16,6 +16,7 @@
 from tests.test_linear_model import main as linear_main
 from tests.test_composite import main as composite_main
 from .test_serialization import main as serialization_main
+from .test_ensemble import main as ensemble_main
```
aqitya, line 19: Reminder to use absolute paths rather than relative paths.

tests/test_ensemble.py
```diff
@@ -0,0 +1,203 @@
...
+            EnsembleModel([m, m, m, 77])
+
+    def test_two_models_identical(self):
+        """
+        this tests that the ensemble model works with two identical model-types with identical states, inputs, etc., one with slightly altered parameters
+        """
+        m = OneInputOneOutputOneEventLM()
+        m2 = OneInputOneOutputOneEventLM(x0={'x1': 2})
+
+        # Make sure they're not the same - 3x the impact
```
aqitya, line 49: maybe stating "they're not the same - 3x their original values" would be a bit clearer

tests/test_ensemble.py
```diff
@@ -0,0 +1,203 @@
...
+        self.assertFalse(em.threshold_met(x1)['x1 == 10'])
+
+        # Transition again
+        x2 = em.next_state(x1, u, 1)
+        # threshold met should be true (because one of 2 models says it is)
+        self.assertTrue(em.threshold_met(x2)['x1 == 10'])
+
+    def test_two_models_different(self):
+        """
+        this tests that the ensemble model works with two different model-types with different states, inputs, etc. Tests how the ensemble model handles the different values
```
aqitya, line 101: This seems redundant... Also may want to consider capitalizing the first letter of the first sentence.

tests/test_ensemble.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        m = OneInputOneOutputOneEventLM()
+        m2 = OneInputOneOutputOneEventAltLM()
+        em = EnsembleModel([m, m2])
+
+        # inputs, states, outputs, events should be a combination of the two models
+        self.assertSetEqual(set(em.inputs), {'u1', 'u2'})
+        self.assertSetEqual(set(em.states), {'x1', 'x2'})
```
aqitya, line 109: Because we use x1 and x2 elsewhere in the code for purposes not only for their states, we should probably either rename our states or rename the later defined variables for x1 and x2.

tests/test_ensemble.py
mstraut: I think the comments could be more consistent in their formatting.

src/prog_models/ensemble_model.py
```diff
@@ -1,6 +1,7 @@
 # Copyright © 2021 United States Government as represented by the Administrator of the
 # National Aeronautics and Space Administration.  All Rights Reserved.
 
+from collections.abc import Iterable
 import numpy as np
 
 from . import PrognosticsModel
```
mstraut, line 7: from prog_models.prognostics_model import PrognosticModel

tests/__main__.py
mstraut: ^ agreed.

mstraut (approved):
I think some more commenting in some areas would be useful and relative paths updated to absolute. Those are minor issues that don't deter me from approving.

src/prog_models/ensemble_model.py
```diff
@@ -30,6 +31,14 @@ class EnsembleModel(PrognosticsModel):
     }
 
     def __init__(self, models, **kwargs):
+        if not isinstance(models, Iterable):
+            raise TypeError(f'EnsembleModel must be initialized with a list of models, got {type(models)}')
```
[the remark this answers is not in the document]
teubert: Generally we expect it to be a list, but I like the idea of supporting any iterable (e.g., tuple). The only way we use it is to iterate through it. 

That said, iterable includes dictionaries, which don't have the behavior we're looking for. So I'm going to change this to sequence.  
image

tests/test_ensemble.py
teubert: I capitalized the first word and added a little more detail on why this test is unique

tests/test_ensemble.py
teubert: I want to keep the states the same so they're consistent with the other linear test models. Instead, I changed the variable names to be x_t0, x_t1, x_t2, ...

aqitya:
Nice PR! 👍

aqitya (approved):

PR #516, opened 2023-04-11
Prog models feature/dict struct

mstraut:
Was using this for something else, but no longer needed. I did happen to do some cleanup and updated all the paths.

### DIFF
--- a/src/prog_models/composite_model.py
+++ b/src/prog_models/composite_model.py
@@ -3,7 +3,7 @@
 
 from collections.abc import Iterable
 
-from prog_models.prognostics_model import PrognosticsModel
+from prog_models import PrognosticsModel
 
 DIVIDER = '.'
 
--- a/src/prog_models/composite_model.py
+++ b/src/prog_models/composite_model.py
@@ -3,7 +3,7 @@
 
 from collections.abc import Iterable
 
-from . import PrognosticsModel
+from prog_models.prognostics_model import PrognosticsModel
 
 DIVIDER = '.'
 
--- a/src/prog_models/models/pneumatic_valve.py
+++ b/src/prog_models/models/pneumatic_valve.py
@@ -5,7 +5,7 @@ from copy import deepcopy
 import numpy as np
 import warnings
 
-from prog_models.prognostics_model import PrognosticsModel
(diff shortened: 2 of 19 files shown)


src/prog_models/composite_model.py
```diff
@@ -3,7 +3,7 @@
 
 from collections.abc import Iterable
 
-from . import PrognosticsModel
+from prog_models.prognostics_model import PrognosticsModel
```
teubert, line 6:
```suggestion
from prog_models import PrognosticsModel
```

This is a good change going from relative to absolute, but we want to import from prog_models without the intermediate method. The same is true for all the others below, as well as Linear and Ensemble models

src/prog_models/models/pneumatic_valve.py
```diff
@@ -26,7 +26,7 @@ def calc_v(x: float, v: float, dv: float, forces: float, Ls: float, new_x: float
     return v + dv
 
 
-class PneumaticValveBase(prognostics_model.PrognosticsModel):
+class PneumaticValveBase(PrognosticsModel):
```
teubert, line 29: Great change, this cleans it up

src/prog_models/prognostics_model.py
```diff
@@ -976,7 +976,7 @@ def check_thresholds(thresholds_met):
         save_pts = config['save_pts']
 
         # confgure optional intermediate printing
-        if config['print']:
+        if config['print']: # BOOMARK FOR MIRYAM
```
teubert, line 979:
```suggestion
        if config['print']:
```
I don't think you meant to check that in

teubert (changes requested):

tests/test_ensemble.py
```diff
@@ -6,7 +6,7 @@
 import sys
 import unittest
 
-from prog_models import EnsembleModel
+from prog_models.ensemble_model import EnsembleModel
```
teubert, line 9:
```suggestion
from prog_models import EnsembleModel
```

teubert (changes requested):
Two remaining comments, but it's almost there

teubert (approved):

PR #519, opened 2023-04-14
Move next_state into utils

teubert:
Small first step towards cleaning up prognostics_model.py. This change moves the state transition wrappers for different integration methods out of prognostics_model.py and into its own file

This change will also help support additional integration methods being worked on by partners.


mstraut (approved):
Looks good. I really appreciate you doing this.

PR #522, opened 2023-04-18
Prog models feature/container none correction

mstraut:
Corrected code that was creating a 3d array in self.matrix

### DIFF
--- a/src/prog_models/utils/containers.py
+++ b/src/prog_models/utils/containers.py
@@ -29,10 +29,18 @@ class DictLikeMatrixWrapper():
                 data = data[np.newaxis].T
             self.matrix = data
         elif isinstance(data, (dict, DictLikeMatrixWrapper)):
-            self.matrix = np.array(
-                [
-                    [data[key]] if key in data else [None] for key in keys
-                ], dtype=np.float64)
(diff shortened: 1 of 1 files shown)


src/prog_models/utils/containers.py
```diff
@@ -29,10 +29,18 @@ def __init__(self, keys: list, data: Union[dict, np.array]):
...
+                self.matrix = np.array(
+                    [
+                        [data[key]] if key in data else [None] for key in keys
+                    ], dtype=np.float64)
+            else:
+                # matrix structure - when data is a dict with arrays as values
+                self.matrix = np.array(
+                    [
+                        data[key] for key in keys if key in data
+                    ], dtype=np.float64)
```
teubert, line 43: Suggested alternate solution
```suggestion
          # ravel is used to prevent vectorized case, where data[key] returns multiple values,  from resulting in a 3D matrix
          self.matrix = np.array(
             [
                  np.ravel([data[key]]) if key in data else [None] for key in keys
              ], dtype=np.float64)
```

teubert (changes requested):
I recommended an alternate solution. Let me know what you think

teubert (approved):

PR #523, opened 2023-04-21
updated import with numpy array to handle dictionaries with values th…

mstraut:
added "from numpy import array" for dictionaries that contain numpy arrays as values.


matteocorbetta (approved):
Approved, since the DictLikeMatrixWrapper will be completely different in a week! :)

PR #524, opened 2023-04-26
calc_error.py housekeeping

aqitya:
Fixing exception and warnings messages as well as reformatting function parameters and documentation.

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -15,6 +15,7 @@ from prog_models.exceptions import ProgModelInputException, ProgModelTypeError, 
 from prog_models.sim_result import SimResult, LazySimResult
 from prog_models.utils import ProgressBar
 from prog_models.utils import calc_error
+from prog_models.utils import containers
 from prog_models.utils.containers import DictLikeMatrixWrapper, InputContainer, OutputContainer
 from prog_models.utils.next_state import euler_next_state, rk4_next_state, euler_next_state_wrapper, rk4_next_state_wrapper
 from prog_models.utils.parameters import PrognosticsModelParameters
@@ -172,21 +173,21 @@ class PrognosticsModel(ABC):
 
         states = self.states
 
-        class StateContainer(DictLikeMatrixWrapper):
+        class StateContainer(containers.StateContainer):
             def __init__(self, data):
                 super().__init__(states, data)
         self.StateContainer = StateContainer
 
         inputs = self.inputs
 
-        class InputContainer(DictLikeMatrixWrapper):
+        class InputContainer(containers.InputContainer):
             def __init__(self, data):
                 super().__init__(inputs, data)
         self.InputContainer = InputContainer
 
         outputs = self.outputs
 
-        class OutputContainer(DictLikeMatrixWrapper):
+        class OutputContainer(containers.OutputContainer):
             def __init__(self, data):
                 super().__init__(outputs, data)
         self.OutputContainer = OutputContainer
@@ -980,13 +981,13 @@ class PrognosticsModel(ABC):
         
         # Auto Container wrapping
         dt0 = next_time(t, x) - t
-        if not isinstance(u, DictLikeMatrixWrapper):
+        if not isinstance(u, InputContainer):
             # Wrapper around the future loading equation
             def load_eqn(t, x):
                 u = future_loading_eqn(t, x)
                 return self.InputContainer(u)
 
-        if not isinstance(self.output(x), DictLikeMatrixWrapper):
+        if not isinstance(self.output(x), OutputContainer):
             # Wrapper around the output equation
             def output(x):
                 # Calculate output, convert to outputcontainer
@@ -1135,7 +1136,7 @@ class PrognosticsModel(ABC):
         raise ProgModelInputException(f"Error method '{method}' not supported")
     
     def estimate_params(self, runs: List[tuple] = None, keys: List[str] = None, times: List[float] = None, inputs: List[InputContainer] = None,
-                        outputs: List[OutputContainer] = None, method = 'nelder-mead', **kwargs) -> None:
+                        outputs: List[OutputContainer] = None, method: str = 'nelder-mead', **kwargs) -> None:
         """Estimate the model parameters given data. Overrides model parameters
 
         Keyword Args:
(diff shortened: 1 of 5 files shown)


src/prog_models/utils/calc_error.py
```diff
@@ -125,18 +127,19 @@ def RMSE(m, times, inputs, outputs, **kwargs):
     return np.sqrt(MSE(m, times, inputs, outputs, **kwargs))
 
 
-def MSE(self, times, inputs, outputs, **kwargs) -> float:
+def MSE(m, times, inputs, outputs, **kwargs) -> float:
```
mstraut, line 130: What was the reason to make this change?

mstraut (changes requested):
The changes look good and make sense, I have a couple of things I'd like to request...

change the import numpy as np to import from numpy.. just what is being used. No need to import all of numpy.

Adding type hints will help clarify the code.
https://docs.python.org/3/library/typing.html

Housekeeping is so important, thanks for taking the time!

src/prog_models/utils/calc_error.py
aqitya: Other calc_error methods have defined the model to be 'm', so for the sake of consistency we changed 'self' to 'm'.

For a similar reason, I might have to include the '-> float' to all the calc_methods or maybe remove the '-> float'.

mstraut: ok, was just trying to understand if there was a reason for it. sounds fine. 

"For a similar reason, I might have to include the '-> float' to all the calc_methods or maybe remove the '-> float'." keep those for clarification. One of my requests for housekeeping is type hints. It's something that's inconsistent in ProgPy and will help future developers when contributing.

kjjarvis (approved):
Nice housekeeping additions, and good change of making the default time step more reasonable

kjjarvis (approved):
Nice housekeeping additions, and good change of making the default time step more reasonable

aqitya:
@mstraut Great suggestion regarding the hints for the functions.

Also, we end up using 8 or so function calls from numpy, so I believe it would probably be best to keep it as it is right now. However, I am not sure where to draw the line between having it explicitly defined in the import statement...

src/prog_models/utils/calc_error.py
```diff
@@ -97,7 +99,7 @@ def MAX_E(m, times, inputs, outputs, **kwargs):
     return err_max
 
 
-def RMSE(m, times, inputs, outputs, **kwargs):
+def RMSE(m, times, inputs, outputs, **kwargs) -> float:
```
mstraut, line 102:
```suggestion
def RMSE(m, times: list[float], inputs: list[dict], outputs: list[dict], **kwargs) -> float:
```

src/prog_models/utils/calc_error.py
mstraut, line 130:
```suggestion
def MSE(m, times: list[float], inputs: list[dict], outputs: list[dict], **kwargs) -> float:
```

src/prog_models/utils/calc_error.py
```diff
@@ -274,16 +280,17 @@ def MAE(m, times, inputs, outputs, **kwargs):
...
-                    warn(f"Model unstable- NaN reached in simulation (t={t})")
+                    warn("Model unstable- NaN reached in simulation (t={})".format(t))
                     break
             err_total += np.sum(
                 np.abs(z.matrix - z_obs.matrix))
             counter += 1
     return err_total/counter
 
-def MAPE(m, times, inputs, outputs, **kwargs):
+def MAPE(m, times, inputs, outputs, **kwargs) -> float:
```
mstraut, line 293:
```suggestion
def MAPE(m, times: list[float], inputs: list[dict], outputs: list[dict], **kwargs) -> float:
```

src/prog_models/utils/calc_error.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 
 
-def MAX_E(m, times, inputs, outputs, **kwargs):
+def MAX_E(m, times, inputs, outputs, **kwargs) -> float:
```
mstraut, line 14:
```suggestion
def MAX_E(m, times: list[float], inputs: list[dict], outputs: list[dict], **kwargs) -> float:
```

src/prog_models/utils/calc_error.py
```diff
@@ -146,64 +149,66 @@ def MSE(self, times, inputs, outputs, **kwargs) -> float:
...
+                    warn("Model unstable- NaN reached in simulation (t={})".format(t))
                     break
-            err_total += np.sum(np.square(z.matrix - z_obs.matrix), where=~np.isnan(z.matrix))
+            err_total += np.sum(np.square(z.matrix - z_obs.matrix), where= ~np.isnan(z.matrix))
             counter += 1
 
     return err_total/counter
 
-def MAE(m, times, inputs, outputs, **kwargs):
+def MAE(m, times, inputs, outputs, **kwargs) -> float:
```
mstraut, line 211:
```suggestion
def MAE(m, times: list[float], inputs: list[dict], outputs: list[dict], **kwargs) -> float:
```

mstraut (changes requested):
Hey, I made some suggestions, for type hints that we're trying to make sure are there for clarity. It takes away some of the confusion for new users and developers.

src/prog_models/prognostics_model.py
```diff
@@ -172,21 +173,21 @@ def __setstate__(self, params: dict) -> None:
 
         states = self.states
 
-        class StateContainer(DictLikeMatrixWrapper):
+        class StateContainer(containers.StateContainer):
```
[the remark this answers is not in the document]
aqitya: Yeah, I think we did, I'll revert the changes back. However, I will still keep enough changes such that the TypeHints remain for calc_error and estimate_params.

src/prog_models/utils/calc_error.py
```diff
@@ -6,12 +6,13 @@
...
 
 from collections.abc import Iterable
+from typing import List
 from warnings import warn
 import math
 import numpy as np
 
 
-def MAX_E(m, times, inputs, outputs, **kwargs):
+def MAX_E(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwargs) -> float:
```
kjjarvis, line 15:
```suggestion
def MAX_E(m, times: List[float], inputs: List[InputContainer], outputs: List[OutputContainer], **kwargs) -> float:
```
Is this consistent with our new notation? If so, let's add it here and for the other methods below

kjjarvis (approved):
Just one last suggestion - this looks great.

PR #525, opened 2023-05-01
Fixing how to initialize StateContainer

aqitya:
When passing in a dictionary to keyword argument 'x0' in any of our given subroutine error methods in calc_error, we are incorrectly transferring the datatype from a dictionary to a StateContainer.


kjjarvis (approved):
Nice bugfix, @aqitya. Do we have an open issue for calc_error tests? If so, let's add a test for this in that issue. If not, let's open one.

mstraut (approved):
Nice job, looks good.

PR #527, opened 2023-05-06
completed: containers, simresult, and test updates

mstraut:
5/5/2023


src/prog_models/data_models/dmd.py
```diff
@@ -420,7 +420,7 @@ def next_state(self, x, u, _):
         if self.B.shape[1] != 0:
            x.matrix += np.matmul(self.B, u.matrix)
         
-        return x   
+        return self.StateContainer(x.matrix)
```
teubert, line 423: Why is this needed? If x has matrix, isn't it already guaranteed to be a state container?

src/prog_models/models/battery_electrochem.py
```diff
@@ -729,7 +730,7 @@ def __init__(self, **kwargs):
         self.param_callbacks['Ro'] = [OverwrittenWarning]
         super().__init__(**kwargs)
 
-    def dx(self, x : dict, u : dict):
+    def dx(self, x: Union[dict, "StateContainer"], u : dict):
```
teubert, line 733:
```suggestion
    def dx(self, x: StateContainer, u : dict):
```

Let's have the type hint only be state container, since that what we want users to use. You can import StateContainer from containers now (as of today). You will need to merge dev first.

tests/test_surrogates.py
```diff
@@ -22,6 +22,7 @@ def tearDown(self):
 
     def test_surrogate_improper_input(self):
         m = ThrownObject()
+        print(m)
```
teubert, line 25:
```suggestion
```

tests/test_pneumatic_valve.py
```diff
@@ -187,6 +187,7 @@ def future_loading(t, x=None):
             self.assertAlmostEqual(x[key], x_test[key], 7)
 
         config = {'dt': 0.01, 'horizon': 800, 'save_freq': 60}
+        print('future_loading\n', future_loading)
```
teubert, line 190:
```suggestion
```

tests/test_linear_model.py
```diff
@@ -432,6 +432,7 @@ def test_linear_pickle(self):
         l = LinearThrownObjectUpdatedInitalizedMethod()
         bytes_l = pickle.dumps(l)
         loaded_l = pickle.loads(bytes_l)
+        print(l == loaded_l)
```
teubert, line 435:
```suggestion
```

tests/test_centrifugal_pump.py
```diff
@@ -158,45 +162,48 @@ def future_loading(t, x=None):
...
+            pump.parameters['wThrust'] = 1e-10
 
     def test_centrifugal_pump(self):
-        self.assertEqual(CentrifugalPump,CentrifugalPumpWithWear)
+        self.assertEqual(CentrifugalPump, CentrifugalPumpWithWear)
         # CentrifugalPump is alias for "with wear"
 
-    def test_centrifugal_pump_namedtuple_access(self):
-        pump = CentrifugalPumpWithWear(process_noise= 0)
+    def test_centrifugal_pump_namedtuple_access(self):  # FAILED
```
teubert, line 183:
```suggestion
    def test_centrifugal_pump_namedtuple_access(self):
```

teubert (commented):
A few cosmetic comments, now I'm going to look into sim_result and container

src/prog_models/models/pneumatic_valve.py
```diff
@@ -6,6 +6,7 @@
 import warnings
 
 from prog_models import PrognosticsModel
+from prog_models.utils.containers import DictLikeMatrixWrapper
```
teubert, line 9:
```suggestion
```

Unused import

src/prog_models/models/centrifugal_pump.py
```diff
@@ -6,6 +6,7 @@
 import warnings
 
 from prog_models import PrognosticsModel
+from prog_models.utils.containers import DictLikeMatrixWrapper
```
teubert, line 9:
```suggestion
```
Unused import

src/prog_models/sim_result.py
```diff
@@ -92,18 +111,18 @@ def remove(self, d : dict = None, t : float = None) -> None:
...
+        self.pop(target_index)
 
     def clear(self) -> None:
         """Clear the SimResult"""
         self.times = []
         self.data = []
+        self._frame = None
 
-    def time(self, index : int) -> float:
+    def get_time(self, index: int) -> float:
```
teubert, line 125: What's the reason for this change? This will change the interface.

src/prog_models/sim_result.py
```diff
@@ -126,13 +145,11 @@ def to_numpy(self, keys = None) -> np.ndarray:
...
-            return np.array([[] for _ in self.data], dtype=np.float64)
-        if isinstance(self.data[0], DictLikeMatrixWrapper) and keys is None:
-            return np.array([u_i.matrix[:, 0] for u_i in self.data], dtype=np.float64)
-        if keys is None:
-            keys = self.data[0].keys()
-        return np.array([[u_i[key] for key in keys] for u_i in self.data], dtype=np.float64)
+        if keys is not None:
+            if all(item in self.frame.columns.to_list() for item in keys) is True:
+                with_keys_numpy = self.frame.drop(['time'], axis=1)[keys].to_numpy(dtype=np.float64)
+                return with_keys_numpy
```
teubert, line 151:
```suggestion
            for key in keys:
                if key not in self.frame:
                    raise KeyError(f'Every key in keys must be present in SimResult. Missing {key}')
            return self.frame[keys].to_numpy(dtype=np.float64)
```

Rewrote to raise exception

src/prog_models/sim_result.py
```diff
@@ -32,7 +33,25 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
...
-    def __eq__(self, other : "SimResult") -> bool:
+    @property
+    def frame(self) -> pd.DataFrame:
+        """
+        Getter -- creating multi row pd.DataFrame from data list of dict
+
+            Return: pd.DataFrame, of all the data and times
+        """
+        if len(self.data) > 0:  #
+            self._frame = pd.concat([
```
teubert, line 44: What is the purpose of _frame as a property instead of a local variable? _frame is recalculated every time you access frame. Are you planning on using caching in a future version.

src/prog_models/utils/containers.py
```diff
@@ -29,14 +30,24 @@ def __init__(self, keys: list, data: Union[dict, np.array]):
...
             raise ProgModelTypeError(f"Data must be a dictionary or numpy array, not {type(data)}")
 
+    @property
+    def frame(self) -> pd.DataFrame:
+        """
+            frame -- Getter for pd.DataFrame
+
+            Returns: pd.DataFrame
+        """
+        self._frame = pd.DataFrame(self.matrix.T, columns=self._keys)
```
teubert, line 48: What is the purpose of _frame as a property instead of a local variable? _frame is recalculated every time you access frame. Are you planning on using caching in a future version.

src/prog_models/utils/containers.py
```diff
@@ -176,6 +190,7 @@ def __repr__(self) -> str:
 
         returns: a string of dictionaries containing all the keys and associated matrix values
         """
-        if len(self.matrix) > 0 and len(self.matrix[0]) == 1:   #the matrix has rows and the first row/list has one value in it
+        if len(self.matrix) > 0 and len(
+                self.matrix[0]) == 1:  # the matrix has rows and the first row/list has one value in it
             return str({key: value[0] for key, value in zip(self._keys, self.matrix)})
-        return str(dict(zip(self._keys, self.matrix)))  
+        return str(dict(zip(self._keys, self.matrix)))
```
teubert, line 196:
```suggestion
        return str(dict(zip(self._keys, self.matrix)))

```
Newline at end of file

teubert (commented):
Great first step towards data frames simresult and container. I'll review the tests next.

tests/test_dict_like_matrix_wrapper.py
```diff
@@ -6,54 +6,52 @@
...
 
     def tearDown(self):
         sys.stdout = sys.__stdout__
-    
+
     def _checks(self, c1):
         self.assertListEqual(c1.keys(), ['a', 'b'])
-        self.assertListEqual(list(c1.values()), [1, 2])
-        self.assertListEqual(list(c1.items()), [('a', 1), ('b', 2)])
+        print(list(c1.values()), [1., 2.])
```
teubert, line 25:
```suggestion
```

tests/test_dict_like_matrix_wrapper.py
```diff
@@ -6,54 +6,52 @@
...
+        self.assertEqual(c1['a'], -1.0)
         # Setting by matrix
-        c1.matrix[1][0] = -2
-        self.assertTrue((c1.matrix == np.array([[-1], [-2]])).all())
-        self.assertEqual(c1['b'], -2)
-
+        c1.matrix[1][0] = -2.0
+        c1.update({'b': -2.0})
+        self.assertTrue(np.all(c1.matrix == np.array([[-1.], [-2.]])))
+        self.assertEqual(c1['b'], -2.0)
```
teubert, line 44:
```suggestion
        c1.matrix[1][0] = -2.0
        self.assertTrue(np.all(c1.matrix == np.array([[-1.], [-2.]])))
        self.assertEqual(c1['b'], -2.0)
        
        # Test update function
        c1.update({'b': -2.0})
        self.assertTrue(np.all(c1.matrix == np.array([[-1.], [-2.]])))
        self.assertEqual(c1['b'], -2.0)
```

The way it's currently done doesn't test .matrix setting. Instead it tests that either matrix or update works.

tests/test_dict_like_matrix_wrapper.py
```diff
@@ -6,54 +6,52 @@
...
+        c1['a'] = -1.0
+        self.assertTrue((c1.matrix == np.array([[-1.], [2.]])).all())
+        self.assertEqual(c1['a'], -1.0)
         # Setting by matrix
-        c1.matrix[1][0] = -2
-        self.assertTrue((c1.matrix == np.array([[-1], [-2]])).all())
-        self.assertEqual(c1['b'], -2)
-
+        c1.matrix[1][0] = -2.0
+        c1.update({'b': -2.0})
```
teubert, line 42:
```suggestion
```
The way it's currently done doesn't test .matrix setting. Instead it tests that either matrix or update works. Update is tested further down

tests/test_dict_like_matrix_wrapper.py
```diff
@@ -6,54 +6,52 @@
...
-        self.assertTrue((c2.matrix == np.array([[-1], [-2]])).all())
-        self.assertEqual(c2['a'], -1)
-        self.assertEqual(c2['b'], -2)
-
+        self.assertTrue(np.all((c2.matrix == np.array([[-1.], [-2.]]))))
+        self.assertEqual(c2['a'], -1.0)
+        self.assertEqual(c2['b'], -2.0)
         # Equality
-        c2 = DictLikeMatrixWrapper(['a', 'b'], {'a': -1, 'b': -2})
+        c2 = DictLikeMatrixWrapper(['a', 'b'], {'a': -1.0, 'b': -2.0})
```
teubert, line 51: Sometime could you show me where the integers became a problem? I'd like to better understand the issue

tests/test_sim_result.py
```diff
@@ -5,102 +5,108 @@
...
 
 from prog_models.models import BatteryElectroChemEOD
 from prog_models.sim_result import SimResult, LazySimResult
 from prog_models.utils.containers import DictLikeMatrixWrapper
 
 
 class TestSimResult(unittest.TestCase):
-    def setUp(self):
-        # set stdout (so it wont print)
+    """def setUp(self):
```
teubert, line 16: Please uncomment the setup/tear down

src/prog_models/data_models/dmd.py
mstraut: fixed as we discussed

teubert (changes requested):
Great initial step towards pandas data frame simresult and *containers. Great work! 

A few small recommendations.

Thanks,
Chris

src/prog_models/sim_result.py
mstraut: No cach in the future, the reason was so that people had the option for the view if they wanted. Since at this stage DF aren't being used in any of the object functions, creating a pandas object every time will just slow down the use of sim_result when it's being forced to convert from data being run without any pandas support. When progpy runs on pandas then it would make sense for sim_result to be fully in pandas as well and no need for the property

src/prog_models/sim_result.py
mstraut: It was so confusing when looking at self.time(x) and self.time[x] side by side that I felt like that change would help with readability. It's also following the python naming convention

src/prog_models/utils/containers.py
mstraut: Chris, you suggested the exact same thing that I have there...

src/prog_models/sim_result.py
teubert: I disagree with your statement about this following python naming conventions though. I think this is more java naming conventions, python prefers to use the name rather than get_name. 

From https://python-consistency.readthedocs.io/en/latest/conventions.html:
"* Do not use get_... Prefer to use the name without the get_ prefix."

That said, I see what you mean about the confusion of self.times[x] and self.time(x). And with times being accessible, this function is also redundant. Why don't we instead move towards making users access times[x] only? We can keep time for one release (per ProgPy policy) with a deprecation warning telling users to use .times[x] instead. What do you think?

src/prog_models/sim_result.py
teubert: We discussed this over teams and decided to keep the generator version for speed. The below change keeps this while implementing the other changes. Namely, 
* Raises an exception if there are missing keys
* removes the unnecessary "is True"
* removes the unnecessary drop

```suggestion
        if keys is not None:
            if not all(item in self.frame.columns.to_list() for item in keys):
                raise KeyError('Every key in keys must be present in SimResult.')
            return self.frame[keys].to_numpy(dtype=np.float64)
```

src/prog_models/data_models/dmd.py
teubert: Didn't we decide to just add a comment here instead of creating a new state container?

src/prog_models/utils/containers.py
```diff
@@ -46,23 +60,23 @@ def __getitem__(self, key: str) -> int:
...
-        index = self._keys.index(key)   #the int value index for the key given
+        index = self._keys.index(key)  # the int value index for the key given
         self.matrix[index] = np.atleast_1d(value)
 
-    
     def __delitem__(self, key: str) -> None:
         """
         removes row associated with key
         """
+        warn('delitem will be deprecated after version 1.5 of ProgPy.', DeprecationWarning, stacklevel=2)
```
teubert, line 79:
```suggestion
```

I don't think this one is being deprecated. del df[key] will still work

src/prog_models/utils/containers.py
```diff
@@ -82,15 +96,19 @@ def __len__(self) -> int:
         """
         returns the length of key list
         """
+        warn('len will be deprecated after version 1.5 of ProgPy.', DeprecationWarning, stacklevel=2)
```
teubert, line 99:
```suggestion
```
I also don't think this will change. len(series) still works

src/prog_models/utils/containers.py
```diff
@@ -82,15 +96,19 @@ def __len__(self) -> int:
...
         return len(self._keys)
 
     def __eq__(self, other: "DictLikeMatrixWrapper") -> bool:
         """
         Compares two DictLikeMatrixWrappers (i.e. *Containers) or a DictLikeMatrixWrapper and a dictionary
         """
-        if isinstance(other, dict): #checks that the list of keys for each matrix match
-            list_key_check = (list(self.keys()) == list(other.keys()))  #checks that the list of keys for each matrix are equal
-            matrix_check = (self.matrix == np.array([[other[key]] for key in self._keys])).all()    #checks to see that each row matches
+        warn('eq will be deprecated after version 1.5 of ProgPy and will be replaced with pd.DataFrame.equals and pd.Series.equals.', DeprecationWarning, stacklevel=2)
```
teubert, line 106:
```suggestion
        warn('Behavior of '==' operator will change after version 1.5 of ProgPy. New behavior will return element wise equality as a new series. To check if two data frames are equals use pd.DataFrame.equals and pd.Series.equals.', DeprecationWarning, stacklevel=2)
```

src/prog_models/utils/containers.py
```diff
@@ -175,9 +196,11 @@ def __repr__(self) -> str:
 
         returns: a string of dictionaries containing all the keys and associated matrix values
         """
-        if len(self.matrix) > 0 and len(self.matrix[0]) == 1:   #the matrix has rows and the first row/list has one value in it
+        warn('DictLikeMatrix will be deprecated after version 1.5 of ProgPy and replaced by Pandas.', DeprecationWarning, stacklevel=2)
```
teubert, line 199:
```suggestion
```

I don't know if we want this message since users won't know they're using DictLikeMatrixWrapper- they'll just continue to use *Container (which will then be a series)

src/prog_models/sim_result.py
```diff
@@ -154,7 +170,7 @@ def plot(self, **kwargs) -> figure:
         Returns:
             Figure
         """
-        return plot_timeseries(self.times, self.data, legend = {'display': True}, options=kwargs)
+        return plot_timeseries(self.times, self.data, legend={'display': True}, options=kwargs)
```
teubert, line 173:
```suggestion
        warn('Behavior of SimResult.plot() will change with version 1.6. New behavior will match that of a pandas data frame.')
        return plot_timeseries(self.times, self.data, legend={'display': True}, options=kwargs)
```

src/prog_models/sim_result.py
```diff
@@ -327,3 +337,21 @@ def data(self) -> List[dict]:
...
+        if len(self.data) > 0:  #
+            self._frame = pd.concat([
+                pd.DataFrame(dict(dframe), index=[0]) for dframe in self.data
+            ], ignore_index=True, axis=0)
+        else:
+            self._frame = pd.DataFrame()
+        if self.times is not None:
+            self._frame.insert(0, "time", self.times)
+            self._frame.reindex()
+        return self._frame
```
teubert, line 357:
```suggestion
```

I don't think this is needed since it's in the parent class (SimResult)

src/prog_models/sim_result.py
```diff
@@ -32,7 +33,23 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
             else:
                 self.data = data
 
-    def __eq__(self, other : "SimResult") -> bool:
+    @property
+    def frame(self) -> pd.DataFrame:
+        """
+            pd.DataFrame: A pandas DataFrame representing the SimResult data
+        """
+        if len(self.data) > 0:  #
```
teubert, line 41:
```suggestion
        warn('frame will be deprecated after version 1.5 of ProgPy.', DeprecationWarning, stacklevel=2)
        if len(self.data) > 0:  #
```

src/prog_models/sim_result.py
```diff
@@ -97,9 +137,24 @@ def pop(self, index: int = -1) -> dict:
         Returns:
             dict: Element Removed
         """
+        warn('pop will be deprecated after version 1.5 of ProgPy. The function will be renamed, popbyindex, and users may begin using it under this name now.',
+            DeprecationWarning, stacklevel=2)
+
         self.times.pop(index)
         return self.data.pop(index)
 
+    def popbyindex(self, index: int = -1) -> dict:
```
teubert, line 146: Should be pop_by_index 

See python naming standards here:https://peps.python.org/pep-0008/#function-and-variable-names

src/prog_models/sim_result.py
```diff
@@ -72,7 +98,21 @@ def index(self, other: dict, *args, **kwargs) -> int:
         Returns:
             int: Index of first sample where other occurs
         """
+        warn('index will be deprecated after version 1.5 of ProgPy. The function will be renamed, indexofdata, and users may begin using it under this name now.',
+            DeprecationWarning, stacklevel=2)
+
         return self.data.index(other, *args, **kwargs)
+    def indexofdata(self, other: dict, *args, **kwargs) -> int:
```
teubert, line 105: Should be index_of_data. See python naming standards: https://peps.python.org/pep-0008/#function-and-variable-names

src/prog_models/sim_result.py
```diff
@@ -34,6 +34,32 @@ def __init__(self, times: list = None, data: list = None, _copy=True):
...
+            'iteration will be deprecated after version 1.5 of ProgPy. The function will be renamed, iterrows, '
+            'and users may begin using it under this name now.',
+            DeprecationWarning, stacklevel=2)
+        super().__iter__(self)
+    
+    def iterrows(self):
+        """
+            Iterates -- through keys
+        """
+        super().__iter__(self)
```
teubert: I actually looked at iterrows, and I think we want to call it itertuples. Iterrows return the index. 

You can also include iterrows and have it return enumerate(self.data)

src/prog_models/sim_result.py
```diff
@@ -34,6 +34,32 @@ def __init__(self, times: list = None, data: list = None, _copy=True):
             else:
                 self.data = data
 
+    def __getitem__(self, item):
+        """
+        created for deprecation warning. [] continues to be handled by parent
+        """
+        warn(
+            '[] will be deprecated after version 1.5 of ProgPy. Users may use pandas Dataframe \'iloc\'  to access '
```
teubert, line 42: We still need to implement iloc

src/prog_models/sim_result.py
```diff
@@ -97,9 +137,24 @@ def pop(self, index: int = -1) -> dict:
...
+    def popbyindex(self, index: int = -1) -> dict:
+        """Remove and return an element
+
+        Args:
+            index (int, optional): Index of element to be removed. Defaults to -1.
+
+        Returns:
+            dict: Element Removed
+        """
+        return self.pop(index)
```
teubert, line 155: This will still raise the deprecation warning when you call pop_by_index. Let's move the logic into this function and have pop call it.

src/prog_models/sim_result.py
```diff
@@ -72,7 +98,21 @@ def index(self, other: dict, *args, **kwargs) -> int:
...
+        """
+        Get the index of the first sample where other occurs
+
+        Args:
+            other (dict)
+
+        Returns:
+            int: Index of first sample where other occurs
+        """
+        return self.index(other, *args, **kwargs)
```
teubert, line 115: Same as pop_by_index. This will still raise the deprecation warning even when they use index_of_data. Move the logic from index into here and have index call this function

src/prog_models/sim_result.py
```diff
@@ -32,7 +34,50 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
...
+                pd.DataFrame(dict(dframe), index=[0]) for dframe in self.data
+            ], ignore_index=True, axis=0)
+        else:
+            self._frame = pd.DataFrame()
+        if self.times is not None:
+            self._frame.insert(0, "time", self.times)
+            self._frame.reindex()
+        return self._frame
+
+    def __eq__(self, other: "SimResult") -> bool:
```
teubert, line 80: Need to implement equals() and add deprecation warning here.

src/prog_models/utils/containers.py
```diff
@@ -19,23 +22,58 @@ def __init__(self, keys: list, data: Union[dict, np.array]):
...
+    def matrix(self, value: np.array) -> None:
+        """
+            matrix -- Setter for numpy array
+
+            Arguments:
+                value -- numpy array
+        """
+        self._matrix = value
+
+    def dot(self, other: Union[np.ndarray, pd.Series, pd.DataFrame]):
```
teubert, line 62: This needs a test

teubert (changes requested):
Looking good. A few recommended changes.

src/prog_models/sim_result.py
```diff
@@ -32,7 +36,56 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
...
+        """
+        created for deprecation warning. [] continues to be handled by parent
+        """
+        warn(
+            '[] will be deprecated after version 1.5 of ProgPy. Users may use pandas Dataframe \'iloc\'  to access '
+            'data by element.',
+            DeprecationWarning, stacklevel=2)
+        return super().__getitem__(item)
+
+    def iloc(self, item):
```
teubert, line 49: We want iloc to work like dataframes to ease transition. That means we need it to work with square brackets. (E.g., result.iloc[10]). 

Let me know if you need some help with this, I have an idea of how to do it

src/prog_models/sim_result.py
```diff
@@ -32,7 +36,56 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
             else:
                 self.data = data
 
-    def __eq__(self, other : "SimResult") -> bool:
+    def __getitem__(self, item):
+        """
+        created for deprecation warning. [] continues to be handled by parent
+        """
+        warn(
+            '[] will be deprecated after version 1.5 of ProgPy. Users may use pandas Dataframe \'iloc\'  to access '
```
teubert, line 44:
```suggestion
            '[] for access by row number will be deprecated after version 1.5 of ProgPy. After v1.5, [] will access by column (e.g., data['state1']), Users may use \'iloc\' to access by row number (e.g., data.iloc[10])'
```

teubert (changes requested):

src/prog_models/sim_result.py
```diff
@@ -32,8 +36,57 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
...
+            'iteration will be deprecated after version 1.5 of ProgPy. The function will be renamed, iterrows, '
+            'and users may begin using it under this name now.',
+            DeprecationWarning, stacklevel=2)
+        return super().__iter__()
+    
+    def iterrows(self):
+        """
+            Iterates -- through keys
+        """
+        return super().__iter__()
```
teubert, line 61: This needs a test

src/prog_models/sim_result.py
```diff
@@ -32,8 +36,57 @@ def __init__(self, times : list = None, data : list = None, _copy = True):
...
+
+    @property
+    def frame(self) -> pd.DataFrame:
```
teubert, line 64: This needs a test

teubert (changes requested):
A few more suggestions

src/prog_models/sim_result.py
```diff
@@ -216,6 +320,7 @@ def __init__(self, fcn : Callable, times : list = None, states : list = None, _c
         """
         self.fcn = fcn
         self.__data = None
+        self.__frame = None
```
teubert, line 323:
```suggestion
```
Unnecessary since it's in parent.

teubert (approved):

mstraut (commented):
Change of warning is fine.

PR #529, opened 2023-05-10
First draft next_state rewrite [Closes #528]

### LINKED ISSUE #528: Integration Method in next_state
**Requirement Text**
Update integration method to apply in next_state instead of simulate_to* (Idea from NGC). Integration method can be set at initialization or afterwards

**Background Information**
Currently, the supported integration methods are euler and rk4. These integration methods are applied in simulation by changing the integration_method argument. They do not apply in next_state, which means they are not applied by particle filters or other methods that use next_state directly. 

**Suggested Solution**
Override next_state method for only discrete models to selected integration method.

**DoD**
edit this to be a list of tasks that need to be completed
 - [ ] Implement feature
 - [ ] Add tests 
 - [ ] update documentation
 - [ ] Add to change notes for next release

teubert:
Update integration logic to apply not just in simulation but also directly to the next_state method. Solution developed in collaboration with partners at NGC

### DIFF
--- a/src/prog_models/prognostics_model.py
+++ b/src/prog_models/prognostics_model.py
@@ -15,7 +15,7 @@ from prog_models.exceptions import ProgModelInputException, ProgModelTypeError, 
 from prog_models.sim_result import SimResult, LazySimResult
 from prog_models.utils import ProgressBar, calc_error
 from prog_models.utils.containers import DictLikeMatrixWrapper
-from prog_models.utils.next_state import euler_next_state, rk4_next_state, euler_next_state_wrapper, rk4_next_state_wrapper
+from prog_models.utils.next_state import next_state_functions
 from prog_models.utils.parameters import PrognosticsModelParameters
 from prog_models.utils.serialization import CustomEncoder, custom_decoder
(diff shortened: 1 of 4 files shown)


MikeAndSpencer (approved):
Looks good! Clever way to implement this. We don't have to do anything for this PR, but I was browsing the centrifugal pump model and it looks like it is considered a discrete model because of the QLeak state but all the other hidden states are continuous. I'm curious if you have any thoughts on how these integration methods could be utilized in a hybrid model, or if it should just be left up to the person writing their model including the integration method in their next state function.

teubert:
> Looks good! Clever way to implement this. You don't have to do anything for this PR with this comment, but I was browsing the centrifugal pump model and it looks like it is considered a discrete model because of the QLeak state but all the other hidden states are continuous. I'm curious if you have any thoughts on how these integration methods could be utilized in a hybrid model, or if it should just be left up to the person writing their model including the integration method in their next state function.

Thanks @MikeAndSpencer - This is a good question. I don't have a good solution now, but I just opened an issue to explore it more.

PR #535, opened 2023-05-26
Linter Cleanup

teubert:
Some cleanup using results form linter. 

Cleanup is to ensure compliance with Python Style Guide. Small things like:
* functions should be in format def f(arg: type = default, arg: type, arg=default) -> type
* Two spaces between code and comment (e.g., code  # comment)
* Removing unused imports
* Dont use import *
* Limiting line length
* etc.

Shouldn't change function of code, just make it more readable.

### DIFF
--- a/examples/sim.py
+++ b/examples/sim.py
@@ -26,7 +26,7 @@ def run_example():
         elif (t < 1800):
             i = 4
         elif (t < 3000):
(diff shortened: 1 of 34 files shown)


teubert:
Skipping review requirements since changes are all in comments/spacing

PR #536, opened 2023-05-27
Scipy integrator

teubert:
Allow any scipy integrator to be passed into integration_method parameter


MikeAndSpencer (approved):
Nice job! I like that you can use all the scipy integrators. I noticed that some of the integrators have different arguments and in the way you implemented this here we just use the default values for each method. I assume it's probably more granular than what we need from this library but I'm curious if you think it's valuable to allow the user to adjust those parameters

teubert:
@MikeAndSpencer configuration can be passed by the 'integrator_config' parameter as a dictionary.

PR #538, opened 2023-05-30
Features/dtw alg

aqitya:
Dynamic Time Warping implementation created as an additional calc_error method along with two nominal test cases.


src/prog_models/utils/calc_error.py
```diff
@@ -7,6 +7,7 @@
 
 from collections.abc import Iterable
 from typing import List
+from fastdtw import *
```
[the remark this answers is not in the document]
teubert: Also, let's move this into the dtw function since it's only needed if you use that function (which most wont). 

I *think* python will only import it once (because it keeps a list in sys.modules), but we should confirm

tests/test_battery.py
teubert, line 1: Here and in other examples you included the benchmarking code. Please remove that (it will be a separate PR)

teubert (changes requested):
It's looking really good. A few suggested changes

tests/test_calc_error.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Consistent results simulated.
+        times = [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0, 6.5, 7.0, 7.5, 8.0, 8.5, 9.0]
+        inputs = [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
+        outputs = [{'x': 2.83}, {'x': 22.868664648480163}, {'x': 40.422881456712254}, {'x': 55.51861687290881}, {'x': 68.06865643702567}, 
+                   {'x': 78.16641111234323}, {'x': 85.89550327176332}, {'x': 91.18868647545982}, {'x': 94.01376508127296}, 
+                   {'x': 94.31711597903195}, {'x': 92.22337598299377}, {'x': 87.67210201789473}, {'x': 80.62858869729064}, {'x': 71.10796509926787}, 
+                   {'x': 59.19579056829866}, {'x': 44.79567793740186}, {'x': 27.97245305860176}, {'x': 8.736607826437163}, {'x': -12.879687324031048}]
+
+        # Has been verified to result in desired values from other DTW algorithms.
```
kjjarvis, line 32:
```suggestion
        # Compare calc_error DTW method to another validated DTW algorithm 
```
Would a comment like this describe what you are doing here?

tests/test_calc_error.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual(m.calc_error(times, inputs, outputs, method = 'dtw'), 4.8146507570483195)
+
+        # Testing case of inconsistent simulation of data (values from times and outputs have been deleted to substantially increase error)
```
kjjarvis, line 35: "Inconsistent simulation of data" doesn't feel quite right here, can we re-word this one too?

tests/test_calc_error.py
```diff
@@ -0,0 +1,81 @@
...
+        times = [0.0, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5.0, 6.5, 7.5, 8.0, 8.5, 9.0]
+        inputs = [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
+        outputs = [{'x': 2.83}, {'x': 40.422881456712254}, {'x': 55.51861687290881}, {'x': 68.06865643702567}, 
+                   {'x': 78.16641111234323}, {'x': 85.89550327176332}, {'x': 91.18868647545982}, {'x': 94.01376508127296}, 
+                   {'x': 94.31711597903195}, {'x': 87.67210201789473}, {'x': 44.79567793740186}, 
+                   {'x': 27.97245305860176}, {'x': 8.736607826437163}, {'x': -12.879687324031048}]
```
kjjarvis, line 41: I was exploring this example a bit and I'm curious to talk more about it. 

Here, 'times' and 'outputs' are two different lengths. The (time, output) pairs are exactly the same as the first dataset in this test (with one point skipped), until about half way through, when the outputs then shift. Is this what was desired with this example? Was the goal to delete the same indices in times and outputs, or different indices? 

If I'm understanding correctly, I think you are trying to create a slightly shifted time series, to show that DTW gives smaller error than MSE. If this is the case, you are achieving that, but I think this example would benefit from some more detailed comments so this is clear. We can talk about it as a group, if you'd like.

kjjarvis (changes requested):
Nice addition of DTW in calc_error! Overall looks great, I just think we can clarify the example a bit

teubert:
* `fastdtw <https://github.com/slaypni/fastdtw>`__ (prog_models only)

   * Requirements met: Dynamic time warping (for dtw metric used in calc_error)
   * Documentation: https://github.com/slaypni/fastdtw
   * Usage Rights: Released under the MIT License
   * Future Support: Unknown- at the time of investigation, the last release was in 3.5 years ago and the last branch was updated 2 years ago. However, the code is simple and the tool is still used by many projects.
   * Other notes: validated by comparing the output of fastdtw to dtaidistance and the algorithm from the 'Towards Data Science' page (https://towardsdatascience.com/dynamic-time-warping-3933f25fcdd). Additionally, the team inspected the output of the package and it seemed reasonable. Finally, a review of open issues on their github repository and a search for known vulnerabilities yielded no concerns. It's only dependency is numpy, which is a trusted package.

tests/test_calc_error.py
aqitya: Yeah, this was exactly what I was going for!

I will add detailed comments to make this clear.

tests/test_calc_error.py
```diff
@@ -0,0 +1,81 @@
...
+    Validating that values are correctly being passed into the new calc_error calls and that we are receiving expected results!
+    """
+
+    def test_DTW(self):
+        """
+        Results from calc_error of DTW work as intended.
+        """
+        m = LinearThrownObject()
+
+        # Consistent results simulated.
```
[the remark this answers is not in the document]
aqitya: # Given preselected data, we are expecting a specific error value from calc_error given method = 'dtw'.
        # By predefining the data, we are removing the 'simulation' step, thus removing any form of variability for the simulated and observed data.
        
        Does this make more sense?

src/prog_models/utils/calc_error.py
teubert, line 1: We wanted to also add times to what's passed to account for inconsistent time deltas between consecutive datapoints (which may be present in collected data)

teubert (approved):
It looks great!

kjjarvis (approved):
This looks great, nice job! I caught just one tiny typo, but otherwise looks ready

PR #539, opened 2023-05-30
Tests/calc error

aqitya:
Moved error checks from each error calculation method to calc_error() to reduce redundancy in code. Furthermore, created tests for calc_error().

Few clarifications of the code below:

Line 1146 of prognostics_model.py follows a format where all three arguments are assumed to be able to take up any one of the acceptable_types: dict, tuple, np.ndarray, list, SimResult, LazySimResult.

check_iterable() ensures that every index of a iterable is either an iterable or Numbers. However, you cannot have both iterables and Numbers exist. Wrote this as a helper function since it may be useful for other functions in prognostics_model.py.


src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,78 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            raise ValueError(len_error)
+        if len(times) < 2:
+            # raise ValueError(f"Must Provide at least 2 data points. Currently only passing in {len(times)}")
+            less_2_error = "Must provide at least 2 data points for times, inputs, and outputs"
+            less_2_error += f" at data location ({_loc})." if _loc is not None else ""
+            raise ValueError(less_2_error)
     
+        x = kwargs.get('x0', self.initialize(inputs[0], outputs[0]))
+        dt = kwargs.get('dt', 1e-99)
+        stability_tol = kwargs.get('stability_tol', 0.95)
```
kjjarvis, line 1164: I'd suggest moving lines 1162-1164 down, right above the checks of these values on 1181, and adding a comment, something like "Checking keyword arguments have appropriate values." It was hard to read in the order it is now.

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,78 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            len_error += f": times = {len(times)}, inputs = {len(inputs)}, outputs = {len(outputs)}"
+            raise ValueError(len_error)
+        if len(times) < 2:
+            # raise ValueError(f"Must Provide at least 2 data points. Currently only passing in {len(times)}")
+            less_2_error = "Must provide at least 2 data points for times, inputs, and outputs"
+            less_2_error += f" at data location ({_loc})." if _loc is not None else ""
+            raise ValueError(less_2_error)
     
+        x = kwargs.get('x0', self.initialize(inputs[0], outputs[0]))
+        dt = kwargs.get('dt', 1e-99)
```
kjjarvis, line 1163: I think this should be 1e99 (not a minus sign). Is that correct?

src/prog_models/utils/calc_error.py
```diff
@@ -37,36 +34,24 @@ def MAX_E(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwar
...
         float: Maximum error between model and data
     """
-    if isinstance(times[0], Iterable):
-        # Calculate error for each
-        error = [MAX_E(t, i, z, **kwargs) for (t, i, z) in zip(times, inputs, outputs)]
-        return max(error)
-
     x = kwargs.get('x0', m.initialize(inputs[0], outputs[0]))
-    dt = kwargs.get('dt', 1e99)
+    dt = kwargs.get('dt', 1e-99)
```
kjjarvis, line 38: I think this should still be 1e99. This change was propagated through all the error methods, and I think it's a mistake. But please confirm, I may be confused. 

Also, the tests currently aren't running to completion due to reaching the time limit. I can't run them on my local machine either. I think it's because the time step for many of the simulations is 1e-99

tests/test_base_models.py
kjjarvis, line 1: Save the profiler stuff for another PR

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+            elif (t < 3000):
+                i = 2
+            else:
+                i = 3
+            return m.InputContainer({'i': i})
+    
+        simulated_results = m.simulate_to(2000, future_loading, **options)
+
+        # Running calc_error before setting incorrect parameters
+        m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt=1)
```
kjjarvis, line 41: Since you don't save this result, I'm not sure it's necessary. This could be a place to save a little computation time

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+from prog_models.models import *
+
+class TestCalcError(unittest.TestCase):
+    """
+    Main Testing class for calc_error.
+
+    Validating that values are correctly being passed into the new calc_error calls and that we are receiving expected results!
+    """
+    def test_calc_error(self):
+        # Note, lowering time steps or increasing simulate threshold may cause this model to not run (takes too long)
```
kjjarvis, line 17:
```suggestion
        # Note: decreasing dt or increasing the time to simulate to will increase computation time
```

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+            str(cm.warning)
+        )
+
+        # Initializing parameters to very erroneous values       
+        m.parameters['qMax'] = 4000
+        keys = ['qMax']
+
+        # Before running estimate_params
+        with self.assertRaises(ValueError):
+            m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt=1)
```
kjjarvis, line 57: Can we add a comment about why this fails? Something like: With these incorrect parameters, the model becomes unstable immediately

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+
+        m.estimate_params([(simulated_results.times, simulated_results.inputs, simulated_results.outputs)], keys, dt=0.5)
+
+        # After running estimate_params. Note that this would not change the outcome of the result
+        with self.assertRaises(ValueError):
+            m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt=1)
```
kjjarvis, line 63: I think I would argue that this doesn't feel like a test of the calc_error functionality. With this highly incorrect qMax value, estimate_params returns the same value as it was initially (i.e. 4000). calc_error has no knowledge of estimate_params, so the call on line 63 is effectively the same as the call on line 57.

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+
+        m.estimate_params([(simulated_results.times, simulated_results.inputs, simulated_results.outputs)], keys, dt=0.5)
+
+        # After running estimate_params. Note that this would not change the outcome of the result
+        with self.assertRaises(ValueError):
+            m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt=1)    
+
+        # Running through various dt values that could work
```
kjjarvis, line 65: Can we make this comment more descriptive? I'm not sure what the point of this test is

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+
+        simulated_results = m.simulate_to(2000, future_loading, **options)
+        m1_sim_results = m1.simulate_to(2000, future_loading, **options)
+
+        data = [(simulated_results.times, simulated_results.inputs, simulated_results.outputs)]
+        data_m1 = [(m1_sim_results.times, m1_sim_results.inputs, m1_sim_results.outputs)]
+
+        # Check out the warnings that are occurring here...
+        # They are being spammed almost. Increasing save_frequency increases spam
+        # Calling estimate_params does not change any of the parameters here because we are always accounting for exceptions...
```
kjjarvis, line 96: Reword for clarity

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+
+        change_params = m1.parameters.copy()
+
+        simulated_results = m.simulate_to(2000, future_loading, **options)
+        m1_sim_results = m1.simulate_to(2000, future_loading, **options)
+
+        data = [(simulated_results.times, simulated_results.inputs, simulated_results.outputs)]
+        data_m1 = [(m1_sim_results.times, m1_sim_results.inputs, m1_sim_results.outputs)]
+
+        # Check out the warnings that are occurring here...
```
kjjarvis, line 94: What is this referring to? I don't see any checks for warnings

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+        # Calling estimate_params does not change any of the parameters here because we are always accounting for exceptions...
+
+        m.estimate_params(data, keys, method='Powell')
+        m1.estimate_params(data_m1, keys, method='CG')
+
+        updated_params = m1.parameters.copy()
+
+        # Checking to make sure estimate_params actually changed values away from the original and to something else
+        self.assertEqual(change_params, updated_params)
+        self.assertNotEqual(orig_params, updated_params)
```
kjjarvis, line 105: This also doesn't really feel like a test of calc_error. But let's chat about it. I'll be curious to hear your perspective!

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+        
+        with self.assertRaises(ValueError):
+            m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt = 1)
+        
+        with self.assertRaises(ValueError):
+            m1.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, dt = 1)
+
+        # Checks to see if stability_tolerance throws error if model goes unstable after threshold
```
kjjarvis, line 113: This is a great test! I like that we can see the error message change as you change stability_tol. That's a great test of our stability_tol functionality

tests/test_calc_error.py
```diff
@@ -0,0 +1,459 @@
...
+        )
+
+        with self.assertRaises(ValueError) as cm:
+            m1.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, 
+                     dt = 1, stability_tol=70)
+        # self.assertEqual(
+        #     'Model unstable- NAN reached in simulation (t=1800.0) before cutoff threshold. '
+        #     'Cutoff threshold is 1900.0, or roughly 95.0% of the data',
+        #     str(cm.exception)
+        # )
```
kjjarvis, line 129:
```suggestion
```
Delete if not using

kjjarvis (commented):
Submitting a partial review for now. Some nice additions to calc_error! I'm looking forward to talking more about these tests.

src/prog_models/prognostics_model.py
```diff
@@ -1075,7 +1075,20 @@ def output(x):
     def __sizeof__(self):
         return getsizeof(self)
 
-    def calc_error(self, times: List[float], inputs: List[InputContainer], outputs: List[OutputContainer], **kwargs) -> float:
+    def check_iterable(self, iter, name, loc = False):
+        """
+        Helper Function to determine if passed in iterable has consistent typings for each index.
+
+        Used to primarily avoid erroneous parameters and for a more descriptive error message than Python's Interpreter.
```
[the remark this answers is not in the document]
aqitya: Hmm... I think 'arguments' makes more sense so for the time being, I will change it to arguments.

src/prog_models/prognostics_model.py
```diff
@@ -1075,7 +1075,20 @@ def output(x):
...
+        """
+        count = sum(isinstance(element, Iterable) and not isinstance(element, (str, bytes)) for element in iter)
+        if 0 < count < len(iter):
+            if loc is not False:
+                raise ValueError(f"Some, but not all elements, are iterable for parameter {name} at data location {loc}")
+            else:
+                raise ValueError(f'Some, but not all elements, are iterable for parameter {name}')
+
+    def calc_error(self, times: List[float], inputs: List[InputContainer], outputs: List[OutputContainer], _loc = None, **kwargs) -> float:
         """Calculate Mean Squared Error (MSE) between simulated and observed
```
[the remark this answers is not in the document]
aqitya: I think it would make sense to make the documentation more generic as well.

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,78 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            len_error += f" at data location ({_loc})" if _loc is not None else ""
+            len_error += f": times = {len(times)}, inputs = {len(inputs)}, outputs = {len(outputs)}"
+            raise ValueError(len_error)
+        if len(times) < 2:
+            # raise ValueError(f"Must Provide at least 2 data points. Currently only passing in {len(times)}")
+            less_2_error = "Must provide at least 2 data points for times, inputs, and outputs"
+            less_2_error += f" at data location ({_loc})." if _loc is not None else ""
+            raise ValueError(less_2_error)
     
+        x = kwargs.get('x0', self.initialize(inputs[0], outputs[0]))
```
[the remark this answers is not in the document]
aqitya: Hmm... it is in line 1195 not as a check, but as an object whose property is being accessed, specifically the name of its type.

The check for the initial state is done in line 1194, which is done by accessing kwargs['x0']. What I can do is replace that section with x instead.

I could also just remove line 1161, but I think it is more consistent with lines 1162 and 1163 to leave it as it is and do the change as mentioned above.

.github/workflows/python-package.yml
```diff
@@ -31,6 +31,16 @@ jobs:
...
+    - name: Upload
+      uses: actions/upload-artifact@v1
+      with:
+        name: output_calls
+        path: output_calls.txt
+    - name: Upload
+      uses: actions/upload-artifact@v1
+      with:
+        name: output_time
+        path: output_time.txt
```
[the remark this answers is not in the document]
aqitya: Sounds good.

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,77 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            'mape': calc_error.MAPE
+        }
+
+        try:
+            method = method_map[method.lower()]
+        except KeyError:
+            # If we get here, method is not supported
+            raise ProgModelInputException(f"Error method '{method}' not supported")
+        
+        # Ensure error methods provide correct results via hand-calculating each one and equating each one.
```
kjjarvis, line 1143: I don't understand this comment. Can we clarify? Happy to help if you want to discuss

src/prog_models/prognostics_model.py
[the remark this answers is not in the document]
kjjarvis: I would suggest changing 1194 back to what it was originally (just with x0 and not x), and delete the current line 1161. Since it's not necessary here, it's (potentially) an added call to the initialize function that is never used. This line has to be called again within each error method, since kwargs (and not x, dt and stability_tol defined in 1161-1163) are passed to each method. That's just my suggestion, happy to hear from you and @teubert as well!

src/prog_models/utils/calc_error.py
```diff
@@ -170,18 +150,11 @@ def MSE(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwargs
     if not isinstance(outputs[0], m.OutputContainer):
         outputs = [m.OutputContainer(z_i) for z_i in outputs]
 
-    # Checks stability_tol is within bounds
-    # Throwing a default after the warning.
-    if stability_tol >= 1 or stability_tol < 0:
-        warn(f"configurable cutoff must be some float value in the domain (0, 1]. "  
-             f"Received {stability_tol}. Resetting value to 0.95")
-        stability_tol = 0.95
```
kjjarvis, line 178: I don't think we can delete this line. In the main calc_error function in prognostics_model, you check if stability_tol is in the bounds and re-define it as 0.95 if not. However, then you pass kwargs, which still has the original stability_tol value, into the actual error method calculation. So without this check here, stability_tol could still be outside of the bounds. 

I think we either need to include this (in all methods), or maybe re-define kwargs in calc_error after the check that it is within bounds.

tests/test_base_models.py
kjjarvis, line 1: We decided to save this for another PR, correct?

kjjarvis (changes requested):
This is looking good, but I found a few things we need to fix (see comments on checks/default values for stability_tol). Happy to work on it together, if you'd like.

tests/test_calc_error.py
aqitya: Yeah, agreed. Will remove these function calls.

src/prog_models/utils/calc_error.py
aqitya: Ah, great catch. I did not see this happening.

Rather than keep these checks, how about I simply change the kwargs['stability_tol'] instead of the other stability_tol that is created? This way we can keep passing in kwargs and we can keep this section of code deleted.

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,77 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            raise TypeError("Times values cannot be strings")
+        if isinstance(times[0], Iterable):
+            # Calculate error for each
+            error = []
+            for r, (t, i, z) in enumerate(zip(times, inputs, outputs)):
+                run_updated = str(r) if _loc is None else _loc + f', {str(r)}'
+                error.append(self.calc_error(t, i, z, _loc=run_updated, **kwargs))
+            return sum(error)/len(error)
+        
+        # Checks stability_tol is within bounds
```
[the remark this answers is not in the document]
kjjarvis: Looks great!

src/prog_models/utils/calc_error.py
kjjarvis: I like this option! Nice fix

tests/test_base_models.py
```diff
@@ -1259,10 +1259,23 @@ def run_tests():
     unittest.main()
     
 def main():
+    import cProfile, pstats
```
kjjarvis, line 1262: Did we decide to leave this for another PR?

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+            m.calc_error(simulated_results.times, simulated_results.inputs, simulated_results.outputs, 
+                     dt = 1, stability_tol=0.7)
+        self.assertEqual(
+            'Model unstable- NaN reached in simulation (t=1800.0)',
+            str(cm.warning)
+        )
+
+    def test_MSE(self):
+        """
+        Base tests that ensure first-level workability within calc_error
```
kjjarvis, line 90: I'm not sure I understand this comment, but it might be software engineering specific language that I'm unfamiliar with

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+        self.assertNotEqual(m.calc_error(results.times, results.inputs, results.outputs),
+                            previous)
+        
+        m2.estimate_params(times = resultsm2.times, inputs = resultsm2.inputs, outputs = resultsm2.outputs, keys = key, dt = 1)
+
+        for i in key:
+            # We can compare with gt because m and m2 originally were the same, and gt was a copy of m.
+            self.assertAlmostEqual(m2.parameters[i], gt[i], 2)
+
+        self.assertNotEqual(m.calc_error(results.times, results.inputs, results.outputs),
```
kjjarvis, line 116: I think this is the same as line 107. m hasn't been changed since then

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+                    previous)
+        
+        self.assertLess(m2.calc_error(resultsm2.times, resultsm2.inputs, resultsm2.outputs), previous)
+
+        # By changing the
```
kjjarvis, line 121: Unfinished comment

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+                    previous)
+        
+        self.assertLess(m2.calc_error(resultsm2.times, resultsm2.inputs, resultsm2.outputs), previous)
+
+        # By changing the 
+        for i in np.arange(0.1, 1, 0.1):
+            m.calc_error(results.times, results.inputs, results.outputs, dt=i)
+
+        for i in range(2, 10):
+            m.calc_error(results.times, results.inputs, results.outputs, dt=i)
```
kjjarvis, line 126: Can we add a comment on the purpose of these tests? It seems like they are just running the calc_error function for different dt values, but not actually saving or comparing the error results. Do we want to add a comparison of some sort?

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+        with self.assertRaises(ValueError) as cm:
+            m.calc_error(incorrectTimes, inputs, outputs)
+        self.assertEqual(
+            "Times, inputs, and outputs must all be the same length. Current lengths: times = 1, inputs = 2, outputs = 2.",
+            str(cm.exception)
+        )
+
+        incorrectTimes = [[0, 1, 2, 4, 5, 6, 7, 8], [0, 1, 2, 3]]
+
+        # Testing when one of the _runs has a data point of different length
```
kjjarvis, line 203:
```suggestion
        # Testing when one of the _runs has an argument of different length
```

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+    
+        # Passing in bool values
+        with self.assertRaises(TypeError) as cm:
+            m.calc_error([False, False, True], [False, False, True], [False, False, True])
+        self.assertEqual(
+            "Data must be a dictionary or numpy array, not <class 'bool'>",
+            str(cm.exception)
+        )
+
+        # Passing in tuples where, where inputs and outputs do not contain dicts or StateContainers, rather contain ints
```
kjjarvis, line 270:
```suggestion
        # Passing in tuples where inputs and outputs do not contain dicts or StateContainers, rather contain ints
```

kjjarvis (commented):
The newest changes look great! Nice fix for the kwargs/stability_tol issue. This PR is nearly ready

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,77 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+        # Throwing a default after the warning.
+        if not isinstance(stability_tol, Number):
+            raise TypeError(f"Keyword argument 'stability_tol' must be either a int, float, or double.")
+        if stability_tol > 1 or stability_tol <= 0:
+            warn(f"Configurable cutoff must be some float value in the domain (0, 1]. Received {stability_tol}. Resetting value to 0.95")
+            stability_tol = 0.95
```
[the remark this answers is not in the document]
kjjarvis: Your fix looks great! I really like the motivation to reduce repeated code, that's great. Your fix by re-defining stability_tol in kwargs solves the problem.

src/prog_models/prognostics_model.py
```diff
@@ -1075,9 +1075,22 @@ def output(x):
     def __sizeof__(self):
         return getsizeof(self)
 
-    def calc_error(self, times: List[float], inputs: List[InputContainer], outputs: List[OutputContainer], **kwargs) -> float:
-        """Calculate Mean Squared Error (MSE) between simulated and observed
+    def check_iterable(self, iter, name, loc = False):
```
teubert, line 1078: This isn't something specific to PrognosticsModel. Could we move this to be its own function in utils.

src/prog_models/prognostics_model.py
```diff
@@ -1075,9 +1075,22 @@ def output(x):
...
+        """
+        Helper Function to determine if passed in iterable has consistent typings for each index.
+
+        Used to primarily avoid erroneous arguments and for a more descriptive error message than Python's Interpreter.
+        """
+        count = sum(isinstance(element, Sequence) for element in iter)
+        if 0 < count < len(iter):
+            if loc is not False:
```
teubert, line 1086:
```suggestion
            if not loc:
```

src/prog_models/prognostics_model.py
```diff
@@ -1075,9 +1075,22 @@ def output(x):
...
         return getsizeof(self)
 
-    def calc_error(self, times: List[float], inputs: List[InputContainer], outputs: List[OutputContainer], **kwargs) -> float:
-        """Calculate Mean Squared Error (MSE) between simulated and observed
+    def check_iterable(self, iter, name, loc = False):
+        """
+        Helper Function to determine if passed in iterable has consistent typings for each index.
+
+        Used to primarily avoid erroneous arguments and for a more descriptive error message than Python's Interpreter.
+        """
```
teubert, line 1083: Add documentation of arguments

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,76 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+            'rmse': calc_error.RMSE,
+            'mae': calc_error.MAE,
+            'mape': calc_error.MAPE
+        }
+
+        try:
+            method = method_map[method.lower()]
+        except KeyError:
+            # If we get here, method is not supported
+            raise ProgModelInputException(f"Error method '{method}' not supported")
```
teubert, line 1141:
```suggestion
            raise KeyError(f"Error method '{method}' not supported")
```
No longer using the custom exceptions for these.

src/prog_models/prognostics_model.py
```diff
@@ -1113,21 +1126,76 @@ def calc_error(self, times: List[float], inputs: List[InputContainer], outputs:
...
+        dt = kwargs.get('dt', 1e99)
+        kwargs['stability_tol'] = kwargs.get('stability_tol', 0.95)
+
+        # Checks stability_tol is within bounds
+        # Throwing a default after the warning.
+        if not isinstance(kwargs['stability_tol'], Number):
+            raise TypeError(f"Keyword argument 'stability_tol' must be either a int, float, or double.")
+        if kwargs['stability_tol'] > 1 or kwargs['stability_tol'] <= 0:
+            warn(f"Configurable cutoff must be some float value in the domain (0, 1]. Received {kwargs['stability_tol']}. Resetting value to 0.95.")
+            kwargs['stability_tol'] = 0.95
```
teubert, line 1185: Let's change this to an exception

src/prog_models/utils/calc_error.py
```diff
@@ -197,16 +170,16 @@ def MSE(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwargs
             # For example, in LSTM models, the first few inputs will not 
             # produce an output until the model has received enough data
             # This is true for any window-based model
-            if any(np.isnan(z_obs.matrix)):
-                if counter < cutoffThreshold:
+            if any (np.isnan(z_obs.matrix)):
```
teubert, line 173:
```suggestion
            if any(np.isnan(z_obs.matrix)):
```
No space in function call

tests/test_base_models.py
teubert: Let's move it to another PR

tests/test_calc_error.py
```diff
@@ -0,0 +1,382 @@
...
+            str(cm.exception)
+        )
+
+        # Test other valid methods do not raise an exception
+        methods = ["max_e", "rmse", "mae", "mape"]
+        for method in methods:
+            m.calc_error(results.times, results.inputs, results.outputs, method=method)
+
+def run_tests():
+    unittest.main()
```
teubert, line 370:
```suggestion
```

We're no longer using the run_test function

teubert (changes requested):

src/prog_models/prognostics_model.py
kjjarvis: I like this idea. It solves our previous issue of needing to re-define kwargs or stability_tol

tests/test_calc_error.py
aqitya: Yeah, the whole purpose of this was to ensure that given some base dt values, our calc_error method would return some value.

In terms of adding some comparisons, should I compare the errors between calc_error calls for a small and large dt value?

kjjarvis: I like the idea of testing that we can input different dt values, good test idea. 

Yes, I would try comparing the errors between two (or more) dt values. Once you've added a few of these checks, I don't think it's necessary to loop over quite so many dt values. Ultimately each iteration in these two loops are doing the same thing, so I think just a couple checks would be sufficient (but correct me if I'm misunderstanding)

aqitya:
Please look over the name choice of the function in the 'input_validation.py' file in the utils directory.

I changed the name from 'check_iterable()' to 'all_none_iterable()'.

tests/test_calc_error.py
aqitya: For the following code segment:
        small = m.calc_error(results.times, results.inputs, results.outputs, dt = 0.001)
        large = m.calc_error(results.times, results.inputs, results.outputs, dt = 0.1
This is true:
         self.assertGreater(small, large)


I was going to write a short description regarding why the larger dt value would result in a lower error, however, wouldn't this behavior be something that is dependent on the specific model we have simulated and the simulated results?

kjjarvis: Hm, I would have expected the opposite actually, that smaller dt gives smaller error. Let me pull your code and take a look

kjjarvis: I don't see any changes to this code

aqitya: Sorry, could you check again. I forgot to push the changes :(

src/prog_models/utils/calc_error.py
```diff
@@ -133,12 +118,12 @@ def MSE(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwargs
...
-        times (list[float]): array of times for each sample
-        inputs (list[dict]): array of input dictionaries where input[x] corresponds to time[x]
-        outputs (list[dict]): array of output dictionaries where output[x] corresponds to time[x]
+        times (list[float], list[list[float]]): Array of times for each sample.
+        inputs (list[dict, SimResult]): Array of input dictionaries where input[x] corresponds to time[x].
+        outputs (list[dict, SimResult]): Array of output dictionaries where output[x] corresponds to time[x].
 
     Keyword Args:
-        x0 (dict, optional): Initial state.
+        x0 (StateContainer, optional): Initial state.
```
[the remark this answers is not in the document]
aqitya: I ended up doing the same for the rest of the methods.

src/prog_models/utils/calc_error.py
```diff
@@ -380,30 +329,25 @@ def DTW(m, times, inputs, outputs, **kwargs):
...
     Keyword Args:
-        x0 (StateContainer): Current State of the model
+        x0 (StateContainer, optional): Initial state.
         dt (float, optional): Maximum time step in simulation. Time step used in simulation is lower of dt and time between samples. Defaults to use time between samples.
         stability_tol (float, optional): Configurable parameter.
             Configurable cutoff value, between 0 and 1, that determines the fraction of the data points for which the model must be stable.
             In some cases, a prognostics model will become unstable under certain conditions, after which point the model can no longer represent behavior. 
             stability_tol represents the fraction of the provided argument `times` that are required to be met in simulation, 
-            before the model goes unstable in order to produce a valid estimate of error. 
+            before the model goes unstable in order to produce a valid estimate of mean squared error.
```
kjjarvis, line 343:
```suggestion
            before the model goes unstable in order to produce a valid estimate of error.
```
I don't think this should say "mean squared error", since that would be specific to MSE but not the other methods. If you agree, this change should be propagated throughout.

src/prog_models/utils/calc_error.py
```diff
@@ -380,30 +329,25 @@ def DTW(m, times, inputs, outputs, **kwargs):
...
 
             If the model goes unstable before stability_tol is met, NaN is returned. 
-            Else, model goes unstable after stability_tol is met, the error calculated from data up to the instability is returned.
+            Else, model goes unstable after stability_tol is met, the mean squared error calculated from data up to the instability is returned.
```
kjjarvis, line 346:
```suggestion
            Else, model goes unstable after stability_tol is met, the error calculated from data up to the instability is returned.
```
Same here, and propagated throughout

tests/test_calc_error.py
```diff
@@ -14,6 +13,360 @@ class TestCalcError(unittest.TestCase):
...
+            [
+                {'x': 1.83},
+                {'x': 36.95},
+                {'x': 62.36},
+                {'x': 77.81},
+            ]]
+        
+        m.calc_error(times, inputs, outputs)
+
+        # Will be testing length errors, will also be providing a 'runs' feedback whenever there are multiple give?
```
kjjarvis, line 192: Reword for clarity

kjjarvis (changes requested):
Almost there! Just need to finish up those dt test changes.

tests/test_calc_error.py
aqitya: Sounds good to me.

src/prog_models/prognostics_model.py
kjjarvis: I don't think this comment was addressed

src/prog_models/utils/calc_error.py
```diff
@@ -133,30 +118,25 @@ def MSE(m, times: List[float], inputs: List[dict], outputs: List[dict], **kwargs
...
         float: Total error
     """
-    if isinstance(times[0], Iterable):
-        # Calculate error for each
-        error = [m.calc_error(t, i, z, **kwargs) for (t, i, z) in zip(times, inputs, outputs)]
-        return sum(error)/len(error)
-
     x = kwargs.get('x0', m.initialize(inputs[0], outputs[0]))
     dt = kwargs.get('dt', 1e99)
     stability_tol = kwargs.get('stability_tol', 0.95)
```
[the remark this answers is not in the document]
kjjarvis: If this is not needed anymore, we should propagate that throughout all of the methods

PR #542, opened 2023-06-01
314 Auto step for piecewise loading profile

teubert:
Two related changes:
1. Added a new loading scheme that defines the logic for piecewise loading profiles
2. Added "stop points". These are points where the simulation will be sure to stop if using "auto" dt. This is important if there are point in the loading profile that you want to capture
3. Will automatically grab the inflection points of the piecewise scheme as stop points (if using "auto" dt)


kjjarvis (approved):
Nice addition of stop_pts. Looks good. I just added a few suggestions for clarity.

Issue #487, opened 2023-01-27
Composite: pm connections


teubert:
**Requirement Text**
Composite model connections should allow for connections between a performance metric and input

**Background Information**
Performance metrics are parameters that are a function of state that are not directly measures. These may affect neighboring systems

**Suggested Solution**
Add connection logic just like the output logic

**DoD**
 - [ ] Implement feature
 - [ ] Add tests

PR #543, opened 2023-06-01
487 Composite: pm connections

teubert:
Extend composite to allow connections from Performance Metrics. This allows the input of one model to be connected to the performance metric from another.

https://github.com/nasa/prog_models/issues/487


aqitya:
Line 199 in composite_model.py reads '# Propogate state', instead of '# Propagate state'.

aqitya:
Having a second, more complex example may be worthwhile to include to the tests.

Maybe a test where we are passing in different models and where circular connections are prevalent as a 'sanity check'.

aqitya:
Overall, nice changes!

It was also a good exercise on my end to see CompositeModels in action!

aqitya (approved):

PR #546, opened 2023-06-06
lazy import

teubert:
Update TF to be imported lazily (only if using LSTM from_data). This will not be used by most users, so this speeds up importing prog_models


mstraut (approved):
Looks good, makes sense.

PR #547, opened 2023-06-06
Switch from Typing to collections.abc

teubert:
abstract base classes in typing are deprecated. Switching to collections.abc.


aqitya (approved):
Looks good!

PR #553, opened 2023-06-14
Float time

teubert:
Update time to use float instead of date time. 

Also, small changes such as:
1.  Moved function to standalone instead of method
2. Removed eta_unix waypoint property
3. Removed vehicle model from use in Trajectory
4. Other cleanup


kjjarvis (approved):
I think this looks great. Impressive debugging, @teubert!!

matteocorbetta (commented):
Looks good!

PR #555, opened 2023-06-14
warn_once

teubert:
Add ability to only throw a warning the first time something occurs. This makes sense for things like Deprecated warnings, that warning the user once is enough. But should not be used for things that are a function of input (e.g., unstable model, input out of range).

This is a quality of life improvement


aqitya:
Honestly, this was something that I wished was implemented, but didn't think it was possible!

aqitya:
Is there any advantages/disadvantages to simply replacing all the warnings with the 'warn_once'?

PR #559, opened 2023-06-20
Add toe and p_met

teubert:
Add Time of Event and Performance Metrics to Template


mstraut (approved):
I remember that Aditya noticed this was missing. Good job, looks good.

PR #561, opened 2023-06-21
Update tutorial

teubert:
Updated tutorial to include additional detail


tutorial.ipynb
```diff
@@ -288,6 +280,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Also, note the the lack of smoothness in the voltage curve. This is limited by the save_freq from sim_cfg. There is a point on the graph every 20 seconds, because that is the frequency at which we save the results. Decreasing the save frequency will result in a cleaner curve.\n",
```
kjjarvis, line 283:
```suggestion
    "Also, note the the lack of smoothness in the voltage curve. This is limited by the save_freq from sim_cfg. There is a point on the graph every 20 seconds, because that is the frequency at which we save the results. Decreasing the save frequency will result in a smoother curve.\n",
```
Just a suggestion

tutorial.ipynb
```diff
@@ -489,6 +485,237 @@
...
+    "\n",
+    "# Plot results\n",
+    "states.plot()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now the curve is much better defined as the simulation progresses. This is frequently of interest since users are mostly interested in behavior close to the event occurance."
```
kjjarvis, line 571:
```suggestion
    "Now the curve is much better defined as the simulation progresses. This is frequently of interest since users are mostly interested in behavior close to the event occurence."
```

tutorial.ipynb
```diff
@@ -489,6 +485,237 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Keep in mind that using a fixed dt, like above, will only save the data at the closest dt increment at or after the save point. So if save_freq is less than dt, save points will be missed. or if save_pts are not a multiple of dt, the resulting point at which data is saved will not be exactly at the save point. Auto step sizes (see next section) can help with this."
```
kjjarvis, line 578:
```suggestion
    "Keep in mind that using a fixed dt, like above, will only save the data at the closest dt increment at or after the save point. So if save_freq is less than dt, save points will be missed. Or if save_pts are not a multiple of dt, the resulting point at which data is saved will not be exactly at the save point. Auto step sizes (see next section) can help with this."
```

tutorial.ipynb
```diff
@@ -489,6 +485,237 @@
...
+   "metadata": {},
+   "source": [
+    "### Integration methods"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Simulation is essentially the process of integrating the model forward with time. By default, simple euler integration is used to propogate the model forward. Advanced users can change the numerical integration method to affect the simulation quality and runtime. This is done using the `integration_method` argument in `simulate_to()`, `simulate_to_threshold()`, or the model parameters. For example:\n",
```
kjjarvis, line 692:
```suggestion
    "Simulation is essentially the process of integrating the model forward with time. By default, simple Euler integration is used to propagate the model forward. Advanced users can change the numerical integration method to affect the simulation quality and runtime. This is done using the `integration_method` argument in `simulate_to()`, `simulate_to_threshold()`, or the model parameters. For example:\n",
```

tutorial.ipynb
```diff
@@ -215,7 +219,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now let's work towards simulating. The following section uses the model created in the \"using the included models\" section\n",
+    "Now let's simulate. The following section uses the model created in the \"using the included models\" section\n",
```
kjjarvis, line 222:
```suggestion
    "Now let's simulate. The following section uses the model created in the \"using the included models\" section.\n",
```

kjjarvis (approved):
Nice additions to the tutorial

PR #567, opened 2023-06-28
Release Review Checklist Items

aqitya:
Complete release review checklist items, namely:
* PrognosticsModel.is_direct_model() -> is_direct() – also everywhere else it’s used
* Add Version note to top of doccomments for all new classes/functions (see release notes) .. versionadded:: 1.5.0
* Rename Chaos example to PCE
* Calc_error: doccomment – x0 as state container, not dict
* test - error_method in estimate params
* In class doccomments for integration_method argument: - next state -> next_state 
* In State estimators doccomments
     * Doccomments for each state estimator. Description of dt should be difference from the last call of .estimate()
     * by default dt is the difference
     * parameters[‘dt’] (missing second quote)


src/prog_models/sim_result.py
```diff
@@ -44,6 +46,8 @@ def __getitem__(self, item):
     
     def __iter__(self):
         """
+        .. versionadded:: 1.5.0
+
```
teubert, line 50:
```suggestion
```
Implemented it in this release to add deprecation warning only- We don't need the message for it

src/prog_models/sim_result.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, times: list = None, data: list = None, _copy=True):
 
     def __getitem__(self, item):
         """
+        .. versionadded:: 1.5.0
+
```
teubert, line 40:
```suggestion
```
Implemented it in this release to add deprecation warning only- We don't need the message for it

src/prog_models/sim_result.py
```diff
@@ -80,13 +88,17 @@ def frame(self) -> pd.DataFrame:
...
+        
         Returns:
             bool: If the value has been calculated
         """
         return self._frame.empty
 
     def __setitem__(self, key, value):
         """
+        .. versionadded:: 1.5.0
+
```
teubert, line 101:
```suggestion
```
Was inherited from UserList before, so it's not new

src/prog_models/sim_result.py
```diff
@@ -96,6 +108,8 @@ def __setitem__(self, key, value):
 
     def __delitem__(self, key):
         """
+        .. versionadded:: 1.5.0
+
```
teubert, line 112:
```suggestion
```
Was inherited from UserList before, so it's not new

src/prog_models/sim_result.py
```diff
@@ -104,6 +118,8 @@ def __delitem__(self, key):
 
     def insert(self, i: int, item) -> None:
         """
+        .. versionadded:: 1.5.0
+
```
teubert, line 122:
```suggestion
```
Was inherited from UserList before, so it's not new

tutorial.ipynb
```diff
@@ -34,7 +34,6 @@
     "* `prog_models.CompositeModel` - model of a system-of-systems, combining multiple interrelated models into a single model\n",
     "* `prog_model.EnsembleModel` - model combining multiple models of the same system into a single model\n",
     "* `prog_models.LinearModel` - parent class for simple linear models\n",
-    "* `prog_models.PrognosticsModel` - parent class for all prognostics models - defines interfaces that a model must implement, and tools for simulating said model \n",
```
teubert, line 37: Why was this removed?

teubert (changes requested):
Great job implementing the review comments. A few small suggestions. Mostly around the version added. There are a few methods that, though you're right that they're newly implemented in this release, I would like to omit because they existed before, just in the parent class

teubert (approved):

PR #573, opened 2023-07-05
fixed some version comments

mstraut:
Fixed the version comments. There were some missing.


teubert (changes requested):
Would it be possible to have this PR without the warn_once changes? Changing an executable line of code, even in a minor way like this would require us to do a hot fix.

teubert (approved):

teubert:
This PR does not change any logical lines of code. Instead, it updates the comments and therefore the documentation generated from these comments. 

Therefore, it is determined that this change does not require a new version/hotfix and it is approved for merge into master.

PR #575, opened 2023-07-07
A few recommended changes

teubert:
I had some suggestions for reworking parameter estimation. Rather than making them all in github, I created this branch. The biggest changes are that I fixed the data for example 1, and that example 2 continues from example 1.

### DIFF
--- a/examples/param_est.ipynb
+++ b/examples/param_est.ipynb
@@ -90,7 +90,7 @@
    "source": [
     "Now we can build a model with a best guess for the parameters.\n",
     "\n",
-    "We will use a guess that our thrower is 20 meters tall and throwing speed of 3.1 m/s and acceleration due to gravity at 15 m/s^2. However, given our times, inputs, and outputs, we can clearly tell this is not true! Let's see if parameter estimation can fix this!"
+    "We will use a guess that our thrower is 20 meters tall, has a throwing speed of 3.1 m/s, and that acceleration due to gravity at 15 m/s^2. However, given our times, inputs, and outputs, we can clearly tell this is not true! Let's see if parameter estimation can fix this!"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


examples/param_est.ipynb
```diff
@@ -74,15 +80,17 @@
...
   },
   {
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Next, we will collect data from the system. Let's pretend we threw the ball once, and collected position measurements."
+    "Now we can build a model with a best guess for the parameters.\n",
+    "\n",
+    "We will use a guess that our thrower is 20 meters tall and throwing speed of 3.1 m/s and acceleration due to gravity at 15 m/s^2. However, given our times, inputs, and outputs, we can clearly tell this is not true! Let's see if parameter estimation can fix this!"
```
kjjarvis, line 93:
```suggestion
    "We will use a guess that our thrower is 20 meters tall, has a throwing speed of 3.1 m/s, and that acceleration due to gravity at 15 m/s^2. However, given our times, inputs, and outputs, we can clearly tell this is not true! Let's see if parameter estimation can fix this!"
```

examples/param_est.ipynb
```diff
@@ -206,75 +202,17 @@
...
-    "keys = ['thrower_height', 'throwing_speed', 'g']"
-   ]
-  },
-  {
-   "attachments": {},
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Then, we'll set the bounds of our values. Let's say that we know our thrower_height will be some value between -15 and 15, our throwing_speed somewhere between 24 and 42, and the gravity is somewhere between -20 and 10.\n",
+    "An additional feature of the `estimate_params()` function is the tolerance feature, or `tol`. The `tol` argument makes our parameter estimation function to continue optimizing until we reach a particular error. This is important for cases where higher model precision is required.\n",
```
kjjarvis, line 205:
```suggestion
    "An additional feature of the `estimate_params()` function is the tolerance feature, or `tol`. The `tol` argument forces our parameter estimation function to continue optimizing until we reach a particular error. This is important for cases where higher model precision is required.\n",
```

examples/param_est.ipynb
```diff
@@ -386,7 +326,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "One thing to note is to have a good estimation of the error, we should be manually measuring the Absolute Mean Error rather than using calc_error().\n",
+    "In this case the error from calc_error is low, but one thing to note is to have a good estimation of the error, we should be manually measuring the Absolute Mean Error rather than using calc_error().\n",
     "\n",
     "The reason being is simple! calc_error() is simulating the error between teh simulated and observed data, however, observed and simulated data in this case are being generated from a model that has noise! In other words, we are comparing the difference of noise to noise, which can lead to inconsistent results!\n",
```
kjjarvis, line 331: Typo in "the" in line 331

kjjarvis, line 331:
```suggestion
    "The reason being is simple! calc_error() is simulating the error between the simulated and observed data, however, observed and simulated data in this case are being generated from a model that has noise! In other words, we are comparing the difference of noise to noise, which can lead to inconsistent results!\n",
```

mrzkp (approved):

PR #576, opened 2023-07-11
Hotfix v1.5.1

teubert:
3 important fixes
  - Change to how dt is added in simulation to prevent floating point errors
  - Fix case where dtype in state from lstm is becomes something other than float
  - Fixes case where hitting horizon in call to time_of_event causes infinite loop
    - previously we only checked that every event was hit.


kjjarvis (approved):
Good catch on a few bugs. Changes look good

kjjarvis:
Approving hotfix - all tests have passed and code has been reviewed

teubert:
Also agree that this hot fix is ready for release

PR #580, opened 2023-07-13
Hotfix v1.5.2

teubert:
Fix issue where LSTM breaks down with new version of tensorflow.

Issue is in the model saving and loading interface, which is optional. I removed this for the hotfix.


kjjarvis (approved):
Nice solution for issue with version of tensorflow

kjjarvis:
Hotfix approved. Nice bugfix and all tests are passing

